PR #6, opened 2016-11-01
Small fix in checkInstallation

simas232:
The error is shown if curSolv is not defined. It is therefore better to
check, whether curSolv is defined rather than checking
whether it is empty.


zertan:
That looks better indeed =D

PR #7, opened 2016-11-02
Simonas branch 1

simas232:
1) Small fix in optimiseProb. To get back Mosek LP solver functionality.
2) Several changes in checkInstallation. See the corresponding commit message for more details.
3) Introduced new non-mandatory fields in model structure: rxnNotes,
rxnReferences, confidenceScores and metCharge. See the corresponding commit message for more details.

### DIFF
--- a/core/addExchangeRxns.m
+++ b/core/addExchangeRxns.m
@@ -21,6 +21,8 @@ function [model addedRxns]=addExchangeRxns(model,reactionType,mets)
 %   Usage: [model addedRxns]=addExchangeRxns(model,reactionType,mets)
 %
 %   Rasmus Agren, 2013-08-01
+%   Simonas Marcisauskas, 2016-11-01 - added support for rxnNotes,
+%   rxnReferences and confidenceScores
 %
 
 if nargin<3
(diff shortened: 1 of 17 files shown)


zertan:
Looks great! We should decide on what to do with the export function. Ideally we would like to have it updated to use libSBML. I will open an issue on it. Which official models use the new annotations?

edkerk:
Using libSBML sounds ideal, it would then also be able to export in FBCv2 format?

simas232:
I agree completely that we should make importModel/exportModel functions to be compatible with libSBML thereby introducing FBC v2 compatibility. Opening a new issue regarding this task seems a perfect idea.

Regarding the utilisation of the new annotation fields, I didn't perform a detailed analysis, but for instance iMK1208 (_Streptomyces coelicolor_) uses all 4 annotation fields (additional data mining was necessary), Yeast 7.6 (_Saccharomyces cerevisiae_) uses metCharge and rxnReferences, iJO1366 (_Escherichia coli_) uses only metCharge. Nevertheless, COBRA now fully supports all these 4 annotation fields, so I expect to see more models in the future, which utilise them.

PR #10, opened 2016-11-10
Small fixes for importModel

edkerk:
Compartment ID of length < 2 called an error after previous pull request.
First attempt to fix this introduced a problem when ID length was 2 or longer, but **not** starting with C_.
Second fix can now deal with comparment IDs like "C_e", "e" and "ext".
Third fix introduces these changes also for compOutside.
(I should probably not push so often...)

### DIFF
--- a/io/importModel.m
+++ b/io/importModel.m
@@ -131,9 +131,9 @@ compartmentMiriams=cell(numel(modelSBML.compartment),1);
 for i=1:numel(modelSBML.compartment)
     compartmentNames{i}=modelSBML.compartment(i).name;
     if isCOBRA==true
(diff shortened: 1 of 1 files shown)


simas232:
Nice work, Ed! I had the issue regarding growing "C_C_C_..'s" before, so it is nice that this issue is fixed. I also think it is fine to pull requests often as long as we fix bugs of existing RAVEN inconsistencies.

zertan:
If we are very serious about these bugs I think the answer is to set up a verification/testing framework. Considering the state master is in now, I'll happily merge anything you find :D.

PR #25, opened 2017-06-05
Deal with NA Pvalues and Fold Changes in reporterMetabolites

muharif:
Hi,

I found two little bugs for the "reporterMetabolites" function:
1. The P-Value and Z-score is showing "NA" in the printed result and result file. This happened due to "NA" in the input P-values, that is causing problem during the usage of "mean" and "std" when correcting for background. "mean" and "std" in Matlab doesn't work well with NA.
2. When geneFoldChanges is provided, it showed "Index exceeds matrix dimensions". It is because there's a mismatch between the size of genes, genesPValues and geneFoldChanges, because genes that are not in model.genes are still there in geneFoldChanges.

I tried to provide small fixes, and it works. Let me know your thoughts.

Thanks!!

### DIFF
--- a/core/reporterMetabolites.m
+++ b/core/reporterMetabolites.m
@@ -64,7 +64,17 @@ end
 
 %Remove the genes which are not in the model
 genePValues(~ismember(genes,model.genes))=[];
+if any(geneFoldChanges)
+    geneFoldChanges(~ismember(genes,model.genes))=[];
+end
 genes(~ismember(genes,model.genes))=[];
+
+%Remove the genes that has NA P-value
+genes(isnan(genePValues))=[];
+if any(geneFoldChanges)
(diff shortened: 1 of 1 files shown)


simas232:
Hi @muharif,

According to my testing, your suggested changes help to resolve the problem with the missing values and fold change discrepancies. Thank you very much for your input!

muharif:
Thanks!

I will use RAVEN a lot in the future, so it benefits me also. Thanks for providing this tools.

Best Regards,
Muhammad Arif

On Jun 6, 2017, at 7:19 PM, Simonas Marcišauskas <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi @muharif<https://github.com/muharif>,

According to my testing, your suggested changes help to resolve the problem with the missing values and fold change discrepancies. Thank you very much for your input!

PR #42, opened 2018-02-09
Fix/import model

edkerk:
~regexp of a cell array does not give required logical (true/false). This was introduced when avoiding the contains() function, not sure if this problem also persists in other RAVEN functions.


simas232:
We used _contains_ function in importModel, importExcelModel (commit bde128188b1c5a6c983222ce755e56e3939cf8fe) and addRxns (commit c9ede76376867d58bb181ce90e110ebe33a18262). I see that we used ~regexp only in importModel, so we are good for now.

PR #45, opened 2018-02-20
feat: put cleanBadChars, cleanBadCharsInModel to 'depreciated' directory

simas232:
In RAVEN1 exportModel did not rely on OutputSBML function found in libSBML. The function generated one long string which was then saved as SBML file. It was therefore necessary to deal with two types of problems: 1) dealing with characters "&, <, >, ", '', %" anywhere in the model; 2) dealing with non-alphanumeric characters in metabolite, reaction, compartment, gene IDs. Both these problems were fixed by cleanBadChars and cleanBadCharsInModel functions. In RAVEN2 exportModel no longer uses any of these functions. In addition, these two functions are not used by any RAVEN function, so they should be discarded from the toolbox. The incompatible characters are now automatically fixed by OutputSBML, which is dependable of exportModel. The non-alphanumeric characters are correspondingly changed into replacement characters, see lines 118-124 in exportModel for more details.


edkerk:
As discussed in https://github.com/SysBioChalmers/RAVEN/commit/49dae94fe20e726b6094fa12ab4ca01ff5527e89, perhaps better to temporarily move the functions into a 'deprecated' folder, which can be cleaned out every now and then?

Hao-Chalmers:
Agree, I thus closed this pull request.

simas232:
@Hao-Chalmers, you should not close this pull request, but request the changes to be done in order to proceed. That is, these two functions should be moved to the new 'depreciated' folder instead of just deleting them.

PR #47, opened 2018-02-23
fix/misc_functions

edkerk:


simas232:
I think we can include one binary parameter, telling the function whether these four files should be contatenated into one. What are your opinions about this?

simas232:
Maybe separated. We shouldn't change the default behaviour of the function.

PR #65, opened 2018-03-20
Fix misc

simas232:
1. Finalized rxnConfidenceScores field type change from cell array of strings to double.
2. Finalized subSystems field type change from cell array of strings to cell array of strings one layer deeper into the field.
3. Revised and updated KEGG reconstruction functions. See more for the correspondent commit for more details. These KEGG functions were tested with the existing KEGG mat files and also with KEGG dump files from KEGG FTP.
4. Other bug fixes.


simas232:
I forgot to mention, that in the following KEGG mat files update, Rhea ids will also be included in rxnMiriams

edkerk:
What do you mean with "performed semi-automatic smart indentation, making things easier for the future development", in commit a3dce33d10457d0feba6c455bd850deee06652aa? Is this something the other developers should/can also implement?

simas232:
I used "smart indent" function, which is available in MATLAB, located in Editor->Indent->Smart. Just made sure that the code is nicely structured with appropriate indentations. Needless to say, the code with appropriate indentations is more amenable for the maintenance and development. I think we need to do exactly the same with the remaining RAVEN functions before publishin RAVEN 2.0 final version.

PR #67, opened 2018-03-27
Fix/raven cobra wrapper

BenjaSanchez:
**Problem this branch addreses:**
When using the COBRA wrapper (`ravenCobraWrapper.m`) with a model fresh out of `reacCbModel.m` from COBRA, the gene rules are not properly converted. This branch fixes that.

**Case study:**
Loading the yeast consensus model with COBRA and looking at the 2nd gene rule (`r_0002`) as a case study, which comes out of COBRA like this:
`(( x(147) & x(243) ) | ( x(147) & x(496) ))`

* **Expected output - output with this fix:**
`(YDL178W and YEL039C) or (YDL178W and YJR048W)`

* **Obtained output in `devel`:**
`(( or x(147) or and or x(243) or ) or | or ( or x(147) or and or x(496) or ))`

It should work with any COBRA model, as the rules are always constructed in the same way, but it would be nice if someone can also try it with a different model just in case.


BenjaSanchez:
@simas232 woops, tiny conflict as I haven't been fetching in a while, but looks easy to solve :)

Issue #69, opened 2018-03-28
fix: deletion of genes in "multiple rules"


BenjaSanchez:
**Description of the issue:**
After a short discussion in issue [#78](https://github.com/SysBioChalmers/YeastMetabolicNetwork-GEM/issues/78) of the yeast repo, I have been looking into how RAVEN handles gene deletions for complexes/isozymes/etc. The results are... well, not good:

- [x] When trying to use `removeGenes.m` to delete genes, the only ones that can be successfully deleted are genes in "single rules", i.e. only one gene in the `grRules` field. Any gene that is involved in a "multiple rule", e.g. `gene1 and gene2`, `gene1 or gene2`, etc. is not deleted by this function.
- [x] Furthermore, the output `reducedModel` looses about half the genes (probably the ones in those multiple rules, but I did not check).
- [x] Finally, `findGeneDeletions.m` does not call `removeGenes.m` (one would expect so to avoid redundant code), and has the major disadvantage that, as stated in the instructions of that function, it disregards complexes (`gene1 and gene2`) and assumes that any "and" is actually an "or". The latter is because it performs deletions by reading the `rxnGeneMat` field, which cannot distinguish between complexes and isozymes (it only has 1s and 0s).

The reader by now will have gathered that all of this is a huge issue when it comes to predicting KOs, and it should be fixed immediately.

**Case study:**
Loading the yeast model and testing `removeGenes.m` with a single rule (`r_0003 <- YAL060W`) and 2 multiple rules (`r_0005 <- YGR032W or YLR342W` and `r_0013 <- YEL038W and YMR009W`).
* **Expected output:**
Each time we use the function we should be able to remove the gene we want. Also, we should observe the following behaviors:
  * Delete `YAL060W` -> delete rxn `r_0003`
  * Delete isozyme `YGR032W` -> don't delete rxn `r_0005` (but change rule to only `YLR342W`)
  * Delete isozymes `YGR032W` & `YLR342W` -> delete rxn `r_0005`
  * Delete subunit `YEL038W` -> delete rxn `r_0013`

* **Obtained output in `devel`:**
The only deletion that actually works is the gene in the single rule (`YAL060W`), all other changes are skipped. The flag `removeRxnsWithComplexes` was tested both =true and =false, with no differences in the output.

* **Reproducing these results:**
[...]

PR #73, opened 2018-04-03
Fix/remove genes

BenjaSanchez:
**Problem this branch addreses:**
Refer to issue #69. It has been partly fixed by adapting `removeGenes.m` to block/not block reactions based on `grRules` and not `rxnGeneMatrix`. It performs the following steps:
* If needed, standardize `grRules` with the function `standardizeGrRules.m`
* Decide if the corresponding reactions can or cannot carry flux with the subfunction `canRxnCarryFlux`, which takes a gene rule and evaluates if the corresponding KO will produce a blocked reaction by **replacing the gene IDs with booleans and assessing the resulting expression**.
* If the gene rule changes in any way, reflect it in the output model, e.g. if an isozyme was deleted from the model, then the corresponding rule will not have that isozyme.

**Case study:**
Loading the yeast model and testing `removeGenes.m` with:
1. A single rule (`r_0003 <- YAL060W`)
2. An isozyme rule (`r_0005 <- YGR032W or YLR342W`)
3. A complex rule (`r_0013 <- YEL038W and YMR009W`)
4. A mixed rule (`r_0002 <- (YDL178W and YEL039C) or (YDL178W and YJR048W)`

* **Expected output - output with this fix:**
Works as it should; reactions get properly blocked (flag `removeBlockedRxns=FALSE`, default) or deleted (flag `removeBlockedRxns=TRUE`) when no gene is left to express for the reaction, and `grRules` updates accordingly in the `reducedModel`.

* **Obtained output in `devel`:**
The only deletion that works is the one for the single rule (`YAL060W`).

* **Reproducing these results:** 
Available in the private repo [gem-scripts](https://github.com/SysBioChalmers/gem-scripts/blob/master/Ben/checkDeletions.m) or [here](https://github.com/SysBioChalmers/RAVEN/files/1872552/checkDeletions.zip).

**Note:** This PR fixes 2/3 to-do's in issue #69. Function `findGeneDeletions.m` has not been adapted yet to use `removeGenes.m`. This will be done in a separate branch.

**Note 2:** The function `ravenCobraWrapper.m` has also been modified to call `standardizeGrRules.m` (both in the direction RAVEN->COBRA and COBRA->RAVEN) and avoid code redundancy. 

**Note 3:** Even though `rxnGeneMatrix` is not used anymore, it is still created by `standardizeGrRules.m` & `ravenCobraWrapper.m` for legacy reasons.

**Note 4:** The available flags for this function have changed. `removeUnusedMets` is still available as first flag, but `removeRxnsWithComplexes` is no longer an available flag, as it should be always be true (can't see a reason to turn it to false now that the function actually works). As function `getModelFromHomology.m` was using the function with the flag set on false, a different output could be expected (but if anything now the results should improve). There are 2 new flags now: `removeBlockedRxns` with a default of false (the behavior of the old version)  and `standardizeRules`  with a default of true (should be skipped when performing intense calculations to avoid too much computation time).

### DIFF
--- a/core/getModelFromHomology.m
+++ b/core/getModelFromHomology.m
@@ -310,7 +310,7 @@ for i=1:numel(models)
     %deleted. This approach works fine for 'and' complexes, but there
     %should be a check that it doesn't keep 'or' genes if it doesn't have
     %to!
-    models{useOrderIndexes(i)}=removeGenes(models{useOrderIndexes(i)},~a,true,true);
+    models{useOrderIndexes(i)}=removeGenes(models{useOrderIndexes(i)},~a,true,true,false);
 end
 
 %Since I want to use mergeModels in the end, I simplify the models further
--- a/core/getModelFromHomology.m
+++ b/core/getModelFromHomology.m
@@ -310,7 +310,7 @@ for i=1:numel(models)
     %deleted. This approach works fine for 'and' complexes, but there
     %should be a check that it doesn't keep 'or' genes if it doesn't have
     %to!
-    models{useOrderIndexes(i)}=removeGenes(models{useOrderIndexes(i)},~a,true,false);
+    models{useOrderIndexes(i)}=removeGenes(models{useOrderIndexes(i)},~a,true,true);
 end
 
 %Since I want to use mergeModels in the end, I simplify the models further
--- a/core/removeGenes.m
+++ b/core/removeGenes.m
@@ -34,7 +34,9 @@ end
 
 %Format grRules and rxnGeneMatrix:
 if standardizeRules
-    model = standardizeGrRules(model);
+    [grRules,rxnGeneMat] = standardizeGrRules(model);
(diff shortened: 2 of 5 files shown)


simas232:
There are several issues to be solved before merging this PR. I will write them here if I will not be able to solve them.

simas232:
I was able to solve importExcelModel problem. Everything is functional now.

simas232:
Just one small request before merging this PR. @BenjaSanchez, RAVEN already has a function `haveFlux`, which identifies reactions that can carry flux. Does the function `canRxnCarryFlux` complement in any way to `haveFlux`? If not, you can use `haveFlux` instead of `canRxnCarryFlux`.

simas232:
I had a check in `canRxnCarryFlux`. This child function definitely complements to `haveFlux`, so it can be retained.

PR #74, opened 2018-04-04
Fix check function uniqueness

simas232:


simas232:
I made the changes. Now the function checks whether the system is Windows or Unix-based and does the path splitting accordingly.

Issue #75, opened 2018-04-06
Static addition of POI to java class path


tpfau:
Coming from #55 I just wanted to test this on my machine and noticed, that RAVEN adds a static java class path file. 
Personally, I think it would be better if this is not done in a static way (the user might move the RAVEN directory, he might want to use a different poi lib etc pp). If you want this to happen by default, you could create a startup.m in the RAVEN main directory that adds the librarys dynamically at startup if the toolbox is on the path or call it in checkInstallation.
This would also avoid the forced matlab restart during setup.

Hao-Chalmers:
Good comments in improving robustness by adding relative (internal) path, other external libs (e.g. libSBML) may also be treated like this (with increased toolbox size though, but convenient for end users).

edkerk:
Could also write a function that checks whether paths are already added or not. If not, then dynamically (?, is that the opposite of static java class path files?) add the POI files.

This function can then be called in the beginning of relevant I/O functions.

Not sure about all other external libs, probably better to keep this in the checkInstallation, as this also ensures that all RAVEN toolbox subfolders are in the Matlab path?

One issues I've been running into is when users do not have admin rights (or at least _enough_ rights), then there can be issues with adding folders and files to path files. Would this also circumvent that?

tpfau:
using `javaaddpath` does not require any specific rights/permissions.  
This is something specific for the POI libraries (or any java librariers that are used).

edkerk:
This is now pushed to master: #89

Issue #55, opened 2018-03-05
Excel I/O problems using MATLAB 2017b and later


danieljcook:
**Edit: a workaround for this problem is provided, see: https://github.com/SysBioChalmers/RAVEN/issues/55#issuecomment-478980100**

Attempting to import an Excel model (.xlsx) using the command "importExcelModel" results in the following error:

Error using importExcelModel (line 140)
Java exception occurred:
java.lang.NoClassDefFoundError: org/apache/commons/collections4/ListValuedMap

	at org.apache.poi.ss.usermodel.WorkbookFactory.create(WorkbookFactory.java:181)

	at org.apache.poi.ss.usermodel.WorkbookFactory.create(WorkbookFactory.java:140)

Caused by: java.lang.ClassNotFoundException:
org.apache.commons.collections4.ListValuedMap

	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)

	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)

	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)

	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)

	... 2 more

There is no problem importing the same model saved as a .xls spreadsheet.

simas232:
This is the problem of Matlab 2017b. It is still possible to fix the Excel model export by adding apache-commons jar file to addJavaPaths, but regarding the Excel file export, onlyl xls format works fine.

According to my investigation, there is nothing bad about Excel I/O functions in RAVEN, but rather the problems of apache-poi itself. I even tried to create a small xlsx file with several rows in columns in one sheet, but the generated xlsx file was non-parsable anyway. Apache POI version 4 should be released soon, let's hope that the upcoming version will fix this issue.

simas232:
By the way, Matlab 2017b is already bundled with apache-poi, version 3.15. I am not sure, how Matlab prioritizes Java packages, if jar files with identical names are found in multiple directories, though.

tpfau:
Out of curiosity:
Does this only happen on Windows or also on unix? 

I recently was annoyed enough to actually code something based on the xlwrite file-exchange submission also using its own bundled POI. 
https://nl.mathworks.com/matlabcentral/fileexchange/65824-xlread

simas232:
The same happens in Unix as well.

tpfau:
Strange, doesn't seem to happen here... 
Do you have a file to test that?

[...]

PR #76, opened 2018-04-06
Making Java library addition dynamic

tpfau:
Just as a suggestion for a solution to #75. 
This change will automatically add the java libraries to the dynamic path is the toolbox is on the path. 
It will no longer require a restart of Matlab at first init and the addition will no longer depend on the actual path of the libraries (as they are added dynamically). 

It might also (potentially) solve #55 , if the issue there was that matlabs POI libraries were used instead of the supplied ones.

### DIFF
--- a/io/addJavaPaths.m
+++ b/io/addJavaPaths.m
@@ -19,28 +19,11 @@ function addJavaPaths()
         fullfile(poiPATH,'stax-api-1.0.1.jar')};
 
     %Open the javaclasspath.txt file or create it
-    %otherwise
-    fid=fopen(fullfile(prefdir,'javaclasspath.txt'),'r');
-    if fid>0
-        current=textscan(fid,'%s','Delimiter','\n');
-        current=current{1};
-        fclose(fid);
-    else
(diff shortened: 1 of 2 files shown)


edkerk:
This would require to run `startup.m` every time MATLAB is restarted, or starting MATLAB in `RAVEN/io`.

Instead, it would be better to call `addJavaPaths` at the beginning of functions that require these libraries (just `exportToExcelFormat` and `importExcelModel` @simas232 ?).

tpfau:
Matlab will run ALL startup.m files which are on the path at startup of matlab.
I.e. if the Raven toolbox is on the path after an initial `checkInstallation` which does `savepath` (and thereby RAVEN/io is on the path) the `startup.m` file is executed.

But yes, adding it to the respective files is an option as well.

edkerk:
Thanks for clarification, your solution works perfectly well! This is IMO a much better solution than adding them statically, or rerunning it every time an Excel I/O function called.

I'll merge this is in 1-2 days in case there are other opinions.

PR #94, opened 2018-04-24
Increase version to rc.2

edkerk:
Note: remember to do this in devel before pushing to master.


edkerk:
I have now clarified in the [Wiki](https://github.com/SysBioChalmers/RAVEN/wiki/DevGuidelines#versioning) that documentation and version.txt should be updated before each PR to `master`.

PR #96, opened 2018-04-26
feat: exportForGit optional file formats

edkerk:
- `exportForGit` can selectively write certain file formats. Default is to write xlsx, xml, yml, txt and mat files.
- make `yml` folder instead of `yaml`


Hao-Chalmers:
The function failed in exporting HMR2, specifically to excel format.

edkerk:
@Hao-Chalmers What is the error message? The function just calls `exportExcelFormat`, so that's probably where a potential bug is?

Hao-Chalmers:
Yeah, it appeared that the bug is from `exportExcelFormat` function.

edkerk:
@Hao-Chalmers But this function doesn't generate/export/process any ComplementaryData or ComplementaryScripts? Or @BenjaSanchez , should the dependencies.txt be in ComplementaryScripts?

@Hao-Chalmers Maybe describe the problem with HMR in an Issue for the `exportToExcelFormat` function?

Hao-Chalmers:
This function does generate ComplementaryData or ComplementaryScripts folders now.

Hao-Chalmers:
Could 'ModelFiles' folder be a better location for this dependencies.txt file? @BenjaSanchez @edkerk

BenjaSanchez:
@edkerk @Hao-Chalmers sounds good! Will move `boundaryMets.txt` there as well. In the afternoon I will test the yeast-GEM with this function for both the saving and releasing functions; will let you know how it goes in the review :)

BenjaSanchez:
@edkerk Changes look good, and nothing gets lost in yeast-GEM. If anything, we win some extra dependencies (libSBML & Matlab), so thank you! Maybe eventually we can transfer most of `increaseVersion.m` as well to RAVEN, but for that some discussion about the optimal Gitflow structure that RAVEN will support should be done first.

My only comment does not have to do with this fix, but I noticed that in line 127 `movefile('*.txt',fullfile(path,'ModelFiles'));` you are moving all `.txt` files, not just `dependencies.txt`. Any specific reason you did this? In the `master` branch of yeast-GEM we also have `version.txt`, which we would not want to move (and would happen when a new release is made, as we now call this function). Can I replace the line for `movefile('dependencies.txt',fullfile(path,'ModelFiles'));`?

edkerk:
No particular reason for using `*.txt`, so fine to replace with the specific filename

PR #97, opened 2018-04-26
feat: addRavenToUserPath

edkerk:
when users has no admin right, it cannot write to `pathdef.m` in matlabroot. Alternatively, this function writes `startup.m` in the userpath, which loads RAVEN to the path every time Matlab is started


Hao-Chalmers:
I've tested this fix.

PR #98, opened 2018-04-27
Fix/findGeneDeletions

edkerk:
Just made this PR to organize comments on solving the `findGeneDeletions` part of this issue: #69 

`findGeneDeletions` should now use `removeGenes` to properly deal with grRules, including isoenzymes.

**This PR should not be merged until tested with a number of different models.**


simas232:
Works fine with my models. PR can be tested further by other before merging.

edkerk:
Also works fine with the models I tested..

PR #100, opened 2018-05-02
Fix ravenCobraWrapper + style fixes

simas232:
-fixed the bug for `ravenCobraWrapper`
-removed uneccesary semicolons after 'end' in several functions
-modifications in comments


edkerk:
@simas232 Function works well, I just clarified the PubChemID related warning. Otherwise, it seems fine to merge.

PR #105, opened 2018-05-14
RAVEN 2.0.0

edkerk:
- fix generation of folders by `exportForGit`
- fix dealing with grRules in model I/O
- rename `addGenes` to `addGenesRaven` for full cobra-compatibility
- remove `version.txt` from `devel` branch, only `master` branch has numbered releases


edkerk:
@simas232 thanks for reminder

Hao-Chalmers:
@edkerk how about release num?

edkerk:
@Hao-Chalmers it's misleading to keep `version.txt` in the `devel` branch, as these (and other development branches) are not actually the version that is mentioned in `version.txt`, but a modification thereof.

Therefore, `devel` branch no longer has this file, while after accepting this PR I will update the `version.txt` in the `master` branch and request a new PR. Accepting this second PR will then result in a release with that version number. Something similar is done for e.g. [`yeast-GEM`](https://github.com/SysBioChalmers/yeast-GEM/commit/2b46934b2a22aff92130e7ce5ee3c361f6c11f2c)

Hao-Chalmers:
Let's wait the documents are made.

PR #113, opened 2018-05-22
feat: compatibility libSBML 5.17.0

edkerk:
- `checkInstallation` confirm that libSBML 5.17.0 is installed (recommended, due to significant speed increase)
- `exportForGit` uses libSBML built-in support to show libSBML version (works from 5.17.0, which is now required for this function. Having this requirement is probably fair, as Git-tracked models should ideally be saved with recent libSBML)

### DIFF
--- a/io/exportForGit.m
+++ b/io/exportForGit.m
@@ -16,7 +16,7 @@ function out=exportForGit(model,prefix,path,formats)
 %
 %   Usage: exportForGit(model,prefix,path,formats)
 %
-%   Eduard Kerkhoven, 2018-05-14
+%   Eduard Kerkhoven, 2018-05-22
 %
 if nargin<4
     formats={'mat', 'txt', 'xlsx', 'xml', 'yml'};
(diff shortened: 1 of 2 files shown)


io/exportForGit.m
```diff
@@ -103,11 +103,12 @@
     disp('COBRA version cannot be found')
 end
 %Retrieve libSBML version:
-fid = fopen('tempModelForLibSBMLversion.xml','w+');
-fclose(fid);
-evalc('[~,~,libSBMLver]=TranslateSBML(''tempModelForLibSBMLversion.xml'',0,0)');
-libSBMLver=libSBMLver.libSBML_version_string;
-delete('tempModelForLibSBMLversion.xml');
+try
+    libSBMLver=OutputSBML;
```
BenjaSanchez: This function requires at least a model as input; the following error occurs otherwise (also in version 5.17.0):
```
Error using OutputSBML
Must supply at least the model as an input argument
USAGE: OutputSBML(SBMLModel, (filename), (exclusiveFlag), (applyUserValidation), (fbcGeneProductOptions))
```

edkerk: @BenjaSanchez it doesn't (as long as an output is specified), see [release notes](http://sbml.org/Software/libSBML/LibSBML_Release_Notes):

- The libSBML-matlab interface function OutputSBML can now be called with an output argument and no inputs to return information about the libSBML version and enabled packages used to build the binary files.

Make sure that your libSBML is 5.17.0, with `which OutputSBML`.

BenjaSanchez: @edkerk you are correct, in my case `OutputSBML` is called from cobratoolbox, as they keep a binary of it as well. That's also why it still takes long to save for me. Hopefully soon they will update it, but in the meantime great that you included the amend :)

PR #115, opened 2018-05-25
merge devel into feat/add_MetaNetX

edkerk:


Hao-Chalmers:
Why is this? It should be done the other way around (from `feature_branch` to `devel`).

edkerk:
To make it up-to-date with devel, while continuing development in this branch with MetaNetX related functions. If it would have been the other way around, the PR would have specified the new functionality.

PR #116, opened 2018-05-25
Revert "merge devel into feat/add_MetaNetX"

Hao-Chalmers:
It should not be this way according to the **Development Guidelines**, which instructed that _development of new features should be in a separate branch from `devel`_ and _make commits only to this branch while developing_ in order to keep compatibility. The PR to `devel` should not be done until _you are happy with your new function/feature_.


edkerk:
Just to clarify for future reference, the feature_branch was updated in #115 with the `devel` branch, to ensure that this feature_branch was also having the latest changes from `devel`. This was also commented as such in that [PR](https://github.com/SysBioChalmers/RAVEN/pull/115#issuecomment-392022325).

In case of suspected mistakes in handling of the repo, it's better to discuss it in e.g. [gitter](https://gitter.im/SysBioChalmers/RAVEN) before reverting someone else's changes.

PR #118, opened 2018-06-13
fix/exportModel and mergeModel; feat/ravenCobraWrapper

edkerk:
- `ravenCobraWrapper` support for more types of metMiriams
- `exportModel` deals with empty (but existing) subSystems field
- `mergeModels` : if the first model does not have metCharges field, it should make a vector of NaN of length(mets) , not a vector of empty cells.

### DIFF
--- a/core/mergeModels.m
+++ b/core/mergeModels.m
@@ -308,8 +308,7 @@ for i=2:numel(models)
             if isfield(model,'metCharges')
                 model.metCharges=[model.metCharges;models{i}.metCharges(metsToAdd)];
             else
-                emptyMetCharge=cell(numel(model.mets)-numel(metsToAdd),1);
-                emptyMetCharge(:)={''};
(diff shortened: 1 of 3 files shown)


edkerk:
@simas232 Good suggestion, will implement this

edkerk:
I implemented the change, the code looks much cleaner, but I don't see a substantial speed improvement (probably because other parts are taking much longer anyway). Also, the code now requires Matlab 2016b due to the use of `contain()` and `join()`.

simas232:
I managed to remove `ravenCobraWrapper` dependence on `contains`, now trying to fix `join`.

edkerk:
@simas232 is this working out, replacing `join`?

simas232:
Yes, it is finally fixed. Could you please check the function before merging?

edkerk:
Works fine here!

PR #154, opened 2018-09-07
fix: importModel extracting grRules from non-FBCv2 models

edkerk:
- commit a390f39c64a695a809990e4767ee1bf2804e600a introduced a bug, where only 'notes' was checked for grRules, and never the 'modifier' field. Now it checks if no grRules is defined yet, and in that case checks in the 'modifier' field.

(note that the diff on GitHub suggests more significant changes. Only lines 518, 519 and 572 were changed)

### DIFF
--- a/io/importModel.m
+++ b/io/importModel.m
@@ -515,60 +515,59 @@ for i=1:numel(modelSBML.reaction)
             %later
             grRules{counter}=geneAssociation;
         end
-    elseif isfield(modelSBML.reaction(i),'modifier')
-        if ~isempty(modelSBML.reaction(i).modifier)
-            rules='';
-            for j=1:numel(modelSBML.reaction(i).modifier)
-                modifier=modelSBML.reaction(i).modifier(j).species;
-                if ~isempty(modifier)
-                    if strcmpi(modifier(1:2),'E_')
-                        index=find(strcmp(modifier,geneIDs));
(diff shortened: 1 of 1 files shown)


edkerk:
@simas232 I'm not sure what bug you're trying to fix, but the suggested change now _doesn't_ import subSystems in FBCv2 models that have the subsystems specified as groups.

Looking at the line of code:
```matlab
if isfield(modelSBML,'groups_group') && any(~cellfun(@isempty,subsystems))
```
`subsystems` is filled with subsystems that were found in the `annotation` field, and this is not the case for FBCv2 models. So the second condition (`any...`) is `false` for FBCv2 models, even though the next block is code is specifically meant for FBCv2 models.

simas232:
Ah, okay. Then it is better to roll back this change. I did such change since I noticed that yeast_7.6.xml from sourceforge, which is in FBCv1, was giving an error. We should prefer FBCv2 than v1, if the issue cannot be resolved.

edkerk:
Inspecting that [SBML file](https://sourceforge.net/projects/yeast/files/yeast_7.6.zip/download), it seems like they used the GROUP package to group identical metabolites that only differ in their localization, so nothing to do with subsystems. I haven't seen this before, I don't think it's common, and I don't see what is the point, so indeed let's revert and deal with this if it occurs with other, FBCv2 models

PR #158, opened 2018-09-24
Style/excelMODELtab

BenjaSanchez:
* Reordered the "MODEL" tab in the excel file to have "TAXONOMY" sooner, as it is more important than other fields.
* Removed default options for author name, email and organization (fields not required by RAVEN)
* Refactored `importExcelModel.m` to simplify switch blocks; they now rely directly on the labels instead of the positions of labels (making it easier for keeping BC in the future)

- [x] If needed, asked first in the [Gitter chat room](https://gitter.im/SysBioChalmers/RAVEN) about this PR


edkerk:
As similarly commented [here](https://github.com/SysBioChalmers/yeast-GEM/pull/163#issuecomment-424245481), taxonomy should refer to the NCBI Taxonomy ID, as valid identifiers.org annotation.

BenjaSanchez:
@edkerk thanks for clarifying! I changed it accordingly. I should also note here that I tested that the i/o excel functions kept functionality without loosing any fields on yeast-GEM.

edkerk:
Yeah, I was worried about that first, also about moving fields around and then importing 'old' models, but the order of fields is not hard-coded in `importExcelModel`. Nice work!

PR #160, opened 2018-09-25
feat: exportForGit writes RAVEN commit id if version unavailable

edkerk:
If RAVEN is not in `master` (where version.txt resides), then it writes the commit id as RAVEN version in the dependencies.txt file

### DIFF
--- a/io/exportForGit.m
+++ b/io/exportForGit.m
@@ -19,7 +19,7 @@ function out=exportForGit(model,prefix,path,formats,masterFlag)
 %
 %   Usage: exportForGit(model,prefix,path,formats,masterFlag)
 %
-%   Benjamin J. Sanchez, 2018-08-03
+%   Benjamin J. Sanchez, 2018-09-26
 %
 if nargin<5
     masterFlag=false;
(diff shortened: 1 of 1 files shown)


io/exportForGit.m
```diff
@@ -51,7 +51,14 @@
     RAVENver = fscanf(fid,'%s');
     fclose(fid);
 catch
-    RAVENver = 'unknown';
+    try
+        currentPath=pwd;
+        cd(RAVENpath);
+        RAVENver = git('log -n 1 --format=%H');
```
BenjaSanchez: How about making this consistent with the COBRA commit, i.e. with the word "commit" and a short hash? If you want I can implement it:
```
RAVEN_toolbox	d9698b2b6db268c63ee3386d5b784a544912549d
COBRA_toolbox	commit d7ce7d6
```

edkerk: Yes! I don't know why I got the longer hash, as I used the same command as for getting the Cobra hash? But if you now how to fix it, then please :)

edkerk:
@BenjaSanchez great, I also tested it here and it works well.

PR #161, opened 2018-09-25
fix: importModel if FBCv2, but groups not used for subsystems

edkerk:
- if a model is FBCv2 and groups seem to be used, it can still be that they don't specify subsystems. So if none of the assigned group reactions are reactionIDs, then skip trying to assign subsystems from groups.


simas232:
What about having Line 667 as `any(idx)`?

edkerk:
You're right, that is more intuitive!

PR #178, opened 2018-10-25
fix: update rxnGeneMat when new genes added

edkerk:
- refactor:
  - `addRxnsGenesMets` should tunnel new genes to `addRxns`, which will then tunnel this to `addGenesRaven`, instead of duplicating code
- fix:
  - `addRxns` should use the existing model if no new genes were found, even if allowNewGenes flag was true
  - `addGenesRaven` should update rxnGeneMat with as many columns as there are new genes


edkerk:
I just encountered a bug, fixing it at the moment.

PR #195, opened 2018-11-23
Fix: addTransport bugfixes

edkerk:
- fix:
  - correct handling of `rxnMiriams` by `addTransport`
  - correct handling of `subSystems` by `addTransport`

### DIFF
--- a/core/addTransport.m
+++ b/core/addTransport.m
@@ -133,9 +133,9 @@ for i=1:numel(toComps)
     if isfield(model,'subSystems')
         ssFiller=filler;
         if isRev==1
-            ssFiller(:)={['Transport between ' fromComp ' and ' toComps{i}]};
+            ssFiller(:)={{['Transport between ' fromComp ' and ' toComps{i}]}};
         else
-            ssFiller(:)={['Transport from ' fromComp ' to ' toComps{i}]};
+            ssFiller(:)={{['Transport from ' fromComp ' to ' toComps{i}]}};
         end
         model.subSystems=[model.subSystems;ssFiller];
     end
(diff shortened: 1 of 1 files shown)


simas232:
Nice fix @edkerk! Just one question, why it is necessary to use double curly brackets in `addTransport`?

edkerk:
@simas232 Because subSystems are stored as arrays of cell arrays, instead of a cell array as it used to be in pre-1.8.0 RAVEN. Another option for that code might be
`ssFiller{:}={['Transport between ' fromComp ' and ' toComps{i}]};`
instead of
`ssFiller(:)={{['Transport between ' fromComp ' and ' toComps{i}]}};`

Hao-Chalmers:
@edkerk the transition of `subSystems` field from _**cell array**_ to _**nested cell array**_ probably happed during **RAVEN** 2.0.0 and 2.1.0. The thing is this transition did broke backward compatibility for many previous models (e.g. quite a lot tINT reconstructions). Allowing _**cell array**_ type `subSystems` is thus preferred, such as by inserting a few lines of detection and conversion at least in `importModel` and `exportModel` functions, and would smooth the adoption process.

edkerk:
@Hao-Chalmers 
This is **not** a recent change, see for instance this commit: 37e0fa5d5d317c73a9176a79973db5846d3c131e from February, This probably broke backwards compatibility if models are stored as `mat` file, but models are routinely shared as SBML file and `importModel` and `exportModel` properly parse the subSystems (see which functions are changed in PR #52). 

So if this is a problem with old `tINIT` generated models that are stored as `.mat`, then these models should be specifically updated, one can write a function to do that, but this does not affect current `importModel` or `exportModel` functions, as the problem here is in the `mat` files, not the SBML files. [COBRA](https://github.com/opencobra/cobratoolbox/blob/master/docs/source/notes/COBRAModelFields.md) also uses cell arrays of cell arrays to deal with multiple subsystems.

PR #199, opened 2018-12-07
Adding potenitally missing comps field

tpfau:
Currently, the `ravenCobraWrapper` requires the comps field to be present in a COBRA model to work. 
This field is an optional field in the current data structure and thus not necessarily present in all models.
As a workaround, I modified the function to create the comps field (based on compartmentIDs present in the metabolite names, as is currently standard practice in COBRA models).

- [ ] Selected `devel` as a target branch

### DIFF
--- a/struct_conversion/ravenCobraWrapper.m
+++ b/struct_conversion/ravenCobraWrapper.m
@@ -272,6 +272,10 @@ else
     %Mandatory RAVEN fields
     newModel.rxns=model.rxns;
     newModel.mets=model.mets;
+    if ~isfield(model,'comps')
+        model.comps = setdiff({''},unique(regexprep(model.mets,'.*\[[^\]]+\]$','')));
+    end
+
     for i=1:numel(model.comps)
         newModel.mets=regexprep(newModel.mets,['\[', model.comps{i}, '\]$'],'');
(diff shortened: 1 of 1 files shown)


BenjaSanchez:
@tpfau this branch was created from `master`, however fixes/new features should always be created from `devel`, as changes to `master` always invoke a new release. I see that you created a branch that follows that (`fixCobraConv`), could you make the PR from that one instead?

tpfau:
Sorry for that, I misunderstood the development guide then, since I assumed, that bugfixes should directly go against master from [here](https://github.com/SysBioChalmers/RAVEN/wiki/DevGuidelines#urgent-bugfix)

BenjaSanchez:
@tpfau no worries, after reading the guidelines I realize that they are indeed confusing. @edkerk @simas232 @Hao-Chalmers have we ever added an urgent bugfix directly to master since the new releasing system? If so, what qualifies as "urgent"? Would that invoke a patch release? The guidelines should be updated to clarify this.

They also state "_For other development (not bugfixes, but for instance introducing new functions or new/updated features for existing functions): make a separate branch from devel [...]_" but I think this is incorrect, normal "non-urgent" bugfixes can (and should) as well go to `devel`.

PR #200, opened 2018-12-07
Adding potenitally missing comps field

tpfau:
Currently, the ravenCobraWrapper requires the comps field to be present in a COBRA model to work.
This field is an optional field in the current data structure and thus not necessarily present in all models.
As a workaround, I modified the function to create the comps field (based on compartmentIDs present in the metabolite names, as is currently standard practice in COBRA models).

### DIFF
--- a/struct_conversion/ravenCobraWrapper.m
+++ b/struct_conversion/ravenCobraWrapper.m
@@ -277,7 +277,7 @@ else
     newModel.rxns=model.rxns;
     newModel.mets=model.mets;
     if ~isfield(model,'comps')
-        model.comps = setdiff({''},unique(regexprep(model.mets,'.*\[[^\]]+\]$','')));
+        model.comps = unique(regexprep(model.mets,'.*\[([^\]]+)\]$','$1'));
     end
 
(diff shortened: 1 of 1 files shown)


struct_conversion/ravenCobraWrapper.m
```diff
@@ -272,6 +272,10 @@
     %Mandatory RAVEN fields
     newModel.rxns=model.rxns;
     newModel.mets=model.mets;
+    if ~isfield(model,'comps')
+        model.comps = setdiff({''},unique(regexprep(model.mets,'.*\[[^\]]+\]$','')));
```
edkerk: This does give the desired result on my PC, what about the following suggestion:

```suggestion
        model.comps = unique(regexprep(model.mets,'.*\[([^\]]+)\]$','$1'));
```

tpfau: You are absolutely right. 
Not sure, what I was thinking there :/

edkerk: :) You should be able to accept the suggested change here on GitHub?

Issue #184, opened 2018-11-06
feat: align RAVEN and COBRA model fields


edkerk:
There are many benefits in having RAVEN functioning as submodule of the [COBRA toolbox](https://github.com/opencobra/cobratoolbox). While RAVEN 2.0 is largely compatible with COBRA through use of the `ravenCobraWrapper` function and using unique function names, a complete alignment of model structure between the two toolboxes would be highly preferred.

To initiate this process, [here](https://doi.org/10.1371/journal.pcbi.1006541.s012) are the definitions of the RAVEN fields.

The main discrepancies are:

1. RAVEN documents compartments in the `metComps` field, instead of detailing this in metabolite IDs
2. RAVEN documents annotations in `metMiriams` and similar fields, instead of `metKEGGID` and similar fields

How should these fields be unified? There is interest from COBRA to move towards the use of the `metComps` field, but what about other discrepancies? What other discrepancies exist?

tpfau:
Hi,
A few things from my point of view:
- `metComps` as a double Vector might be faster but is more prone to model modifications leading to out of sync fields (I would personally prefer a cell array of strings representing the ID of the compartment instead of the position in the comps vector).
- `rxnMiriams` and similar: I'm missing a qualifier field in the struct. Also, is this a struct array or a single struct with 2 fields. (If the latter, wouldn't a  struct array make more sense?)
- `rev` We decided to drop this field mainly because it was used as an indicator for flux reversibility, which is goverened by the flux bounds. I think we could bring it back as an annotation type of information on whether a reaction could be reversible physiologically (e.g. according to data derived from eQuilibrator or similar). But I would really want to avoid this field to indicate flux reversibility, as that information is otherwise redundant and already present in `lb`.
- `rxnComps` This field (same as `geneComps`) looks odd to me. Reactions can have multiple compartments (e.g. transport), and those are derived from the participating metabolites, and for genes, while yes, their gene product could be localised to a specific compartment, this should be represented by different GPR rules for reactions in different compartments (or am I missing something here)?

tpfau:
Oh and one more thing:
We recently introduced fields for additional Variables/Constraints to prepare for concepts like Enzyme Capacity constraints (as provided by GECKO). Those fields hold information which is separate from Metabolites/Reactions but interacts with those. 
[...]

PR #202, opened 2018-12-14
fix: constructEquations

Hao-Chalmers:
- fix: get the reversibility information by querying both `lb` and `ub` fields, instead of the obsolete `rev` field that could be inaccurate or unavailable
- doc: refine function description with `Input` and `Output` sub-headings
- feat: additional option is added for allowing the generation of equations with met formulas

**NOTE:** None of these changes affect backward-compatibility of any of the modified functions. All reactions in a model should be organized in their forward direction, typically by setting `ub = 1000` and `lb = -1000 (or 0)`, so that their equations can be correctly constructed by the `constructEquations` function.

### DIFF
--- a/core/constructEquations.m
+++ b/core/constructEquations.m
@@ -72,7 +72,7 @@ for i=1:numel(Rindexes)
     end
     %Define stoich coeffs and reversibility:
     stoichCoeffs = model.S(Mindexes,Rindexes(i));
-    isrev        = model.rev(Rindexes(i))==1;
+    isrev        = model.lb(Rindexes(i))<0;
     
     %Construct equation:
     equationStrings{i} = buildEquation(mets,stoichCoeffs,isrev);
--- a/core/constructEquations.m
+++ b/core/constructEquations.m
@@ -74,7 +74,7 @@ for i=1:numel(Rindexes)
     end
     %Define stoich coeffs and reversibility:
     stoichCoeffs = model.S(Mindexes,Rindexes(i));
-    isrev        = model.lb(Rindexes(i))<0;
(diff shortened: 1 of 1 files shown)


edkerk:
It makes a lot of sense to stop using the `rev` field, but it needs to be mentioned explicitely in the documentation that the `lb` fields is queried for determining reversibility. There can be cases where `ub = 0` and `lb = -1000`, is this strictly reversible? True, the reaction should then have been defined in the other direction, but it can be good to mention this. 

Something like:

> All reactions are defined in their forward direction, while the `lb` field is queried to determine whether reactions can also support a flux in reverse direction.

edkerk:
Also, in the process of abandoning the use of the `rev` field, the functionality of `convertToIrrev` should be changed, as this currently uses the `rev` field to identify what reactions are reversible.

BenjaSanchez:
@edkerk based on a previous discussion in https://github.com/SysBioChalmers/RAVEN/issues/184#issuecomment-436653399, will the `rev` vector still be able to be read by `convertToIrrev` with some `readRevVector=TRUE` flag? As GECKO uses `convertToIrrev` as one of the first steps, would be nice to know so I can handle the change later.

Hao-Chalmers:
Plan to add an option for constructing reactions using `metformulas`, this appears to be quite useful when balancing reaction.

edkerk:
> How about having an option of constructing reactions using `metFormulas`, this appears to be quite useful when balancing reaction.

Sure! I'm not sure when this is useful, but if you find a use for it that's great. You've implemented it correctly, not breaking previous functionality, so I have no problem with this part.

PR #203, opened 2019-01-08
fix: getKEGGModelForOrganism error report bug

simas232:
- During model reconstruction from KEGG, the users may download pre-trained HMMs ZIP file, but sometimes MATLAB may fail to extract it. Upon the fix, such error is now correctly reported to the user. This is done by checking existence for _K00844.hmm_ (hexokinase), which is present in all HMM sets
- Misc formatting fixes

### DIFF
--- a/external/kegg/getGenesFromKEGG.m
+++ b/external/kegg/getGenesFromKEGG.m
@@ -2,16 +2,17 @@ function model=getGenesFromKEGG(keggPath,koList)
 % getGenesFromKEGG
 %   Retrieves information on all genes stored in KEGG database
 %
+%   Input:
 %   keggPath	if keggGenes.mat is not in the RAVEN\external\kegg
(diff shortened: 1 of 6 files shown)


simas232:
Also added `Input`/`Output` to description section for all KEGG functions

Hao-Chalmers:
The check for existence of `K00844.hmm` looks good. How about adjust the error message to something like "The HMMs files seem improperly extracted, please ...".

simas232:
What about this? "The HMM files seem improperly extracted (not available in dataDir/hmms). Please remove dataDir directory and re-run the same command". This would not print "dataDir", but the actual `dataDir` value.

Hao-Chalmers:
The parenthesis appear redundant, simply put "The HMM files seem improperly extracted and not found in ...".

How about take way  It's a bit strange to include in error messages

PR #222, opened 2019-05-20
feat: setExchangeBounds new function to set exchange bounds for given mets

JonathanRob:
New function `setMetExchange` allows users to set the exchange flux bounds for a given list of metabolites. This should be helpful for specifying e.g., media conditions, or other nutrient uptake/export fluxes.

### DIFF
--- a/core/setMetExchange.m
+++ b/core/setMetExchange.m
@@ -1,0 +1,197 @@
+function [exchModel,unusedMets] = setMetExchange(model,mets,lb,ub,closeOthers,mediaOnly)
+% setMetExchange
+%   Define the exchange flux bounds for a set of metabolites.
+%
+%   model         a model structure
+%   mets          a cell array of metabolite names (case insensitive) or 
+%                 metabolite IDs, or a vector of metabolite indices
+%                 (opt, default all exchanged metabolites)
+%   lb            lower bound of exchange flux. Can be either a vector of
+%                 bounds corresponding to each of the provided metabolites,
+%                 or a single value that will be applied to all.
+%                 (opt, default -1000)
+%   ub            upper bound of exchange flux. Can be either a vector of
+%                 bounds corresponding to each of the provided metabolites,
+%                 or a single value that will be applied to all.
+%                 (opt, default 1000)
+%   closeOthers   close (set lb=ub=0) exchange reactions corresponding to
+%                 all other exchanged metabolites not present in the
+%                 provided list
+%                 (opt, default true)
+%   mediaOnly     only consider exchange reactions involving exchange to or
+%                 from the extracellular (media) compartment. Reactions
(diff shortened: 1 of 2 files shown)


Hao-Chalmers:
two quick comments:
1. how about rename to `setExchangeMets`.
2. now RAVEN starts to include subheadings `Inputs:` and `Outputs:`, please add them to the header

core/setMetExchange.m
```diff
@@ -0,0 +1,212 @@
...
+
+
+% handle input arguments
+if nargin < 2
+    mets = [];
+elseif ischar(mets)
+    mets = {mets};  % in case only one metabolite is provided as a string
+end 
+if nargin < 3 || isempty(lb)
+    lb = -1000;
```
edkerk: Instead of using -1000 and 1000 as default LB and UB, models can have different default values set under `model.annotation.defaultLB` and `model.annotation.defaultUB`. So if now lb or ub is defined when running this function, first check if default bounds are defined for the model, while only as last resort use -1000 and 1000 as default values.

JonathanRob: Oh, nice, I didn't know that was a possible field. I've now modified the function to check for this field before defaulting to +/-1000.

JonathanRob:
@Hao-Chalmers I added the subheadings to the function header. 

As for the suggested name change to `setExchangeMets`, I don't mind changing it to this, but I'm not sure it's as accurate. `setExchangeMets` sounds to me as if I'm defining which metabolites are exchange metabolites, which is a property of the S-matrix. I named it `setMetExchange` because I am setting the metabolite exchange rate for a given set of metabolites.

I'm actually a bit uncertain as to which name I prefer, and am open to other suggestions.

Hao-Chalmers:
@JonathanRob how about `setExchangeFluxs ` or `setExchangeRates `.

edkerk:
@Hao-Chalmers @JonathanRob what about `setExchangeBounds`?

Hao-Chalmers:
`setExchangeBounds ` seems more fit

JonathanRob:
I agree, `setExchangeBounds` seems to be the most appropriate.

Hao-Chalmers:
@JonathanRob is it feasible to have an associated `test` function to evaluate/test the performance of this function?

@edkerk it might be good to include a `test` folder allowing the addition of testing functions in the future?

JonathanRob:
@Hao-Chalmers Of course, I can certainly arrange such a function if there is a place for it.

edkerk:
> is it feasible to have an associated `test` function to evaluate/test the performance of this function?

@Hao-Chalmers Do you mean tests that check that the various RAVEN functions are working as intended, as quality control for future development? This would be really good, although it is probably quite time intensive. I would then prioritize #184 instead.

PR #223, opened 2019-05-20
fix: mergeModels

ckitti:

- fix bug where `mergeModels` would incorrectly modify the S matrix if different metabolites share the same metabolite names (`model.metNames`). Based on standard that corresponding metabolites across different compartments have unique metabolite IDs (`model.mets`), function updated to match metabolites by IDs instead of names
- updated comments and documentation


simas232:
Hmm, I think this issue can by solved in different, more strategic way. If there are different metabolites with the same `metNames` in the same compartment, then these metabolites would be stacked together during `exportToExcelFormat` -> `importExcelModel` pipeline. My suggestion is that we should modify `checkModelStruct` instead. An additional check for overlapping `metNames` in the same compartment would be added, which would yield a warning. What are your opinions @edkerk, @Hao-Chalmers?

edkerk:
@ckitti: Instead of changing `mergeModels` functionality from merge by name to merge by ID, add this as a parameter, so that the user can choose to _either_ merge by ID _or_ name, as each of these functions can be appropriate in different situations.

@simas232: Good idea to have `checkModelStruct` include a check for finding identical metabolite names in the same compartment with different IDs. However, this is separate from the functionality of `mergeModels`.

ckitti:
* [X]  include options to **either** merge by _id_ **or** _name_
* [X]  define parameter in `mergeModels` to choose which of the above modes should be run
* [X]  clarify the two different modes in the documentation at the beginning of the function.

`mergeModels` updated with additional argument for parameter `metParam`, which designates whether the function matches by `model.metNames` (metabolite names)  or `model.mets` (IDs). To ensure backwards compatibility, function matches by name if argument is unspecified. Documentation has been updated to reflect changes and how to choose between modes.

Is there a better way to delineate the script following the two different modes to make it more streamlined and efficient?

simas232:
@edkerk, we should consider calling `checkModelStruct` in the beginning of `mergeModels` function. I mean, the modified `checkModelStruct` function including the new check for overlapping `metNames` in one compartment.

If we conclude that `checkModelStruct` is not necessary in `mergeModels`, we should still include the checks for overlapping `mets` and `metNames` in the latter function.

simas232:
@ckitti, I have one weird suggestion for the sake of the code efficiency and simplicity. Maybe we don't need `metParam` parameter at all.

I suggest that the function `mergeModels` should firstly check for overlapping `metNames`. Depending on the test result, the function would then proceed with metabolites matching by `metNames` or `mets`. The warning would be printed if `metNames` were not unique in the same compartment.

ckitti:
@simas232 That seems the most elegant solution. In that case in which the function is carried out semi-automatically, I suppose including `checkModelStruct` would be imperative to proof the structure before initiating. In the case that a model erroneously has duplicate met IDs and was not checked manually by user before merging, a warning produced by conflicting `model.mets` would enable user to pinpoint issue more easily.

simas232:
@ckitti, even though we include `checkModelStruct` in `mergeModels`, an additional check for overlapping `metNames` in one compartment would be needed. Nonetheless, this change does not significantly slow `mergeModels` function, so it would be fine.

edkerk:
> @ckitti, I have one weird suggestion for the sake of the code efficiency and simplicity. Maybe we don't need `metParam` parameter at all.
> 
> I suggest that the function `mergeModels` should firstly check for overlapping `metNames`. Depending on the test result, the function would then proceed with metabolites matching by `metNames` or `mets`. The warning would be printed if `metNames` were not unique in the same compartment.

Aha, now I get what you meant. Interesting idea. Are there situations where `metNames` don't match, but models should still not be merged by `mets`?

simas232:
@edkerk, I think that metabolites matching by `mets` is the safest way to merge the multiple GEMs. Even if the same metabolite IDs are used for actually different compounds across the GEMs, such discrepancies are solved by considering such metabolites as conflicting and renaming their corresponding IDs.

However, such metabolites matching by `mets` comes at its own cost. That is, if one merges e.g. three bacterial GEMs and we have cytosolic and extracellular oxygen in all the models, the resulting merged GEM would have six separate `mets` for oxygen. So the exhaustive manual curation would be needed to merge oxygen and other metabolites.

It seems we need to re-think how to identify the same metabolites between different GEMs. Metabolites matching by `metNames` is still preferable way to merge GEMs in the most of the cases. The Sco4, for which such merging does not work appropriately, has several compounds in the same compartment under the different `mets`, but identical `metNames`, e.g. (R)-3-Hydroxydecanoyl-[acyl-carrier protein]. So maybe we should include the check for `metNames` uniqueness in the same compartment and rename all such occurences. For instance, (R)-3-Hydroxydecanoyl-[acyl-carrier protein], which occurs twice in Sco4 would be renamed as "(R)-3-Hydroxydecanoyl-[acyl-carrier protein] (3hdecACP_c)" and "(R)-3-Hydroxydecanoyl-[acyl-carrier protein] (3hdecACPredQ_c)"

ckitti:
@simas232 I was under the impression that metabolite IDs would act as the universal standard between various formats of GEMs. For example, I recently worked with another GEM that did not specify metabolite names at all but used metabolite IDs. As for COBRA, the equations are constructed using the IDs. 

As for the caveat you mentioned regarding introducing duplicates of metabolites, I think I may have addressed this in how I wrote the function to find the unique new metabolites between the main and additional models (please see lines 427-433). Although I must mention I have not yet tested with merging more than 2 models, so there may have some faults there - will test later today.

simas232:
@ckitti, I see your point for metabolite IDs as the universal standard, but there is no such a common consensus in modelling community, although the databases like BiGG or MetaNetX try to address this. The lack of such consensus is the key reason why e.g. water under different metabolite IDs between the models would be recognised as different metabolites in `mergeModels.`

Okay, I am looking forward to the testing results. You can try to merge two Sco4 models upon using `mets` matching.

edkerk:
@ckitti @simas232: what is the way forward here? I think it's probably easiest and most versatile to have `metParam` as parameter (with `metNames` as default to keep backwards compatibility).
@ckitti, when adding a new parameter, make sure that other functions that call `mergeModels` also parse the correct parameter value (again, `metNames` for backwards compatibility, unless you specifically want to change the behaviour of the other function):
`getModelFromHomology`, `fillGaps`, `fitTasks`, `getINITModel`, `tutorial2018`, `linkMetaCycKEGGRxns`,  `copyToComps`

simas232:
@edkerk, I do support the implemention of `metParam` parameter. However, I still think that we should add the check for the overlapping `metNames` in one compartment. This check would not report all the issues with the models, but only the first occurence. No adjustments would be made to `metNames`, we would leave this up to the user.

edkerk:
@simas232 OK, but let's implement that in a separate PR, so we can merge this one already.

ckitti:
@edkerk Issue has been fixed for all remaining functions that use `mergeModel` in their scripts with exception of script `tutorial2018`, in which user can see explicitly what the input is for the use of `mergeModel` in this case.

PR #226, opened 2019-06-30
fix: misc gap-filling, addTransport, exportModel, getModelFromHomology bugs

edkerk:
- fix:
  - slightly decrease tolerance when using gurobi solver, particularly when solving MILP. If tolerance is set not strict enough, `fillGaps` does not function well.
  - `addTransport` when defining new reaction IDs
  - error message in `addRxnsGenesMets`
  - bug in `getModelFromHomology` when using multiple template models
  - bug in `getModelFromHomology` where genes were partial matched: YAL001W is not YAL001W-A
  - bug in `fillGaps` where all genes from template models were kept, not only the ones annotated to gap-filling reactions
  - `exportModel` replace illegal characters in grRules
  - bug in `addTransport` when adding multiple transport reactions
  - `addMets`: multiple compartments can be given as either rows or columns
- feat:
  - `addTransport` has list of added reaction IDs as alternative 

### DIFF
--- a/core/getModelFromHomology.m
+++ b/core/getModelFromHomology.m
@@ -390,7 +390,7 @@ if ~isempty(preferredOrder) && numel(models)>1
         
         %Remove all the genes that were already found and add the other
         %ones to allUsedGenes
-        models{useOrderIndexes(i)}=removeGenes(models{useOrderIndexes(i)},allGenes{i+1}(genesToDelete),true,false,false);
+        models{useOrderIndexes(i)}=removeGenes(models{useOrderIndexes(i)},allGenes{i+1}(genesToDelete),true,true,false);
         allUsedGenes(usedGenes)=true;
         
         %Remove the deleted genes from finalMappings and allGenes.
--- a/core/getModelFromHomology.m
+++ b/core/getModelFromHomology.m
@@ -456,7 +456,7 @@ for i=1:numel(models)
                     repString=[repString ') or (' fullGeneList{b(l)}];
                 end
                 %Use regexprep instead of strrep to prevent partial matches
-                models{useOrderIndexes(i)}.grRules{j}=regexprep(models{useOrderIndexes(i)}.grRules{j},[geneName{1} '($|\s|\))'],[repString '$1']);
+                models{useOrderIndexes(i)}.grRules{j}=regexprep(models{useOrderIndexes(i)}.grRules{j},['(^|\s|\()' geneName{1} '($|\s|\))'],['$1' repString '$2']);
             else
                 %Then search in the non-replaceable genes. There could only
                 %be one match here
--- a/core/getModelFromHomology.m
+++ b/core/getModelFromHomology.m
@@ -455,7 +455,8 @@ for i=1:numel(models)
                 for l=2:numel(b)
                     repString=[repString ') or (' fullGeneList{b(l)}];
                 end
-                models{useOrderIndexes(i)}.grRules{j}=strrep(models{useOrderIndexes(i)}.grRules{j},geneName{1},repString);
+                %Use regexprep instead of strrep to prevent partial matches
+                models{useOrderIndexes(i)}.grRules{j}=regexprep(models{useOrderIndexes(i)}.grRules{j},[geneName{1} '($|\s)'],[repString '$1']);
             else
                 %Then search in the non-replaceable genes. There could only
                 %be one match here
--- a/core/getModelFromHomology.m
+++ b/core/getModelFromHomology.m
@@ -456,7 +456,7 @@ for i=1:numel(models)
                     repString=[repString ') or (' fullGeneList{b(l)}];
                 end
                 %Use regexprep instead of strrep to prevent partial matches
-                models{useOrderIndexes(i)}.grRules{j}=regexprep(models{useOrderIndexes(i)}.grRules{j},[geneName{1} '($|\s)'],[repString '$1']);
+                models{useOrderIndexes(i)}.grRules{j}=regexprep(models{useOrderIndexes(i)}.grRules{j},[geneName{1} '($|\s|\))'],[repString '$1']);
             else
                 %Then search in the non-replaceable genes. There could only
                 %be one match here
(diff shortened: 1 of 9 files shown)


edkerk:
@Hao-Chalmers You're right, it's actually reducing the tolerance... The point is that with the previous values, `fillGaps` seems to have found a solution to its MILP problem, but the reported solution is actually not functional. By _reducing_ the tolerance, the MILP solver will not return such false positive solutions, although the calculation time probably increases slightly.

JonathanRob:
I noticed this type of bug related to fillGaps (through the use of tINIT) a while ago, but couldn't figure out the source, as it was very difficult to reproduce. I'm glad you pinpointed the cause, and will test it out on the problematic case I had before to see if it solves the problem.

core/getModelFromHomology.m
```diff
@@ -455,7 +455,8 @@
                 for l=2:numel(b)
                     repString=[repString ') or (' fullGeneList{b(l)}];
                 end
-                models{useOrderIndexes(i)}.grRules{j}=strrep(models{useOrderIndexes(i)}.grRules{j},geneName{1},repString);
+                %Use regexprep instead of strrep to prevent partial matches
+                models{useOrderIndexes(i)}.grRules{j}=regexprep(models{useOrderIndexes(i)}.grRules{j},[geneName{1} '($|\s|\))'],[repString '$1']);
```
JonathanRob: This fix will prevent partial matches for genes with the same starting names, e.g. YAL001W and YAL001W-A, as you claimed. However, it will not prevent partial matches for genes with the same *ending* names, e.g., AGENE and GENE. I would therefore suggest the following line of code:

`models{useOrderIndexes(i)}.grRules{j}=regexprep(models{useOrderIndexes(i)}.grRules{j},['(^|\s|\()' geneName{1} '($|\s|\))'],['$1' repString '$2]);`

io/exportModel.m
```diff
@@ -113,7 +113,13 @@ function exportModel(model,fileName,exportGeneComplexes,supressWarnings)
 model.mets=regexprep(model.mets,'([^0-9_a-zA-Z])','__${num2str($1+0)}__');
 model.comps=regexprep(model.comps,'([^0-9_a-zA-Z])','__${num2str($1+0)}__');
 if isfield(model,'genes')
-    model.genes=regexprep(model.genes,'([^0-9_a-zA-Z])','__${num2str($1+0)}__');
+    problemGenes=find(~cellfun('isempty',regexp(model.genes,'([^0-9_a-zA-Z])')));
+    originalGenes=model.genes(problemGenes);
+    replacedGenes=regexprep(model.genes(problemGenes),'([^0-9_a-zA-Z])','__${num2str($1+0)}__');
+    model.genes(problemGenes)=replacedGenes;
+    for i=1:numel(problemGenes)
+        model.grRules=strrep(model.grRules,originalGenes{i},replacedGenes{i});
```
JonathanRob: Similar situation here. You are using `strrep` to replace gene names/IDs, which will cause problems with partial matches. I would suggest the following:

`model.grRules = regexprep(model.grRules, ['(^|\s|\()' originalGenes{i} '($|\s|\))'], ['$1' replacedGenes{i} '$2']);`

Hao-Chalmers:
@JonathanRob very good. Would also be nice to include your testing code for confirming the fixes to `fillGaps` into this PR. I recommend to have this script either in the exiting `tutorial` folder or a new `test` folder.

core/getModelFromHomology.m
edkerk: Thanks @JonathanRob, you're completely right.
When suggesting a code change when reviewing a PR, see [here](https://help.github.com/en/articles/reviewing-proposed-changes-in-a-pull-request) point 6, this makes is very convenient to apply the suggested changes.

JonathanRob:
@Hao-Chalmers I agree that it would be useful to include my code for testing the `fillGaps` fix. However, this code involves the use of a modified version of a published GEM, a modified version of a task list, and running tINIT. All of these files would need to be provided, in addition to the cell line RNA-Seq expression profiles, and the run time is several hours on a standard computer.

In addition to all of this, I cannot confirm that the error will reproduce with my code. As I mentioned above, I had difficulty reproducing this error, which I why I was unable to identify the source myself.

This is why my "test" should only be treated as an observation that the tolerance adjustment seems to resolve a problem related to my very specific case (**so far**), but is far from a proper test.

Hao-Chalmers:
@edkerk it could be outside of this PR and I agree to start an issue regarding the preparation of tests.

@JonathanRob

core/getModelFromHomology.m
JonathanRob: Ah, very nice, I'll use this next time. This code-commenting feature is still a bit new to me.

ckitti:
I have noticed when I am using `gapFill` function (on up-to-date `devel` branch) that the function is adding duplicate reactions from template model to draft model but having different representations in `model.S` matrix. By that I mean if we used `constructEquations` on these reactions, the printed equations are different (see table). Has anyone encountered this issue before or know what may be causing this?

ID | NAME | EQUATION | EC-NUMBER | GENE ASSOCIATION | LOWER BOUND | UPPER BOUND
-- | -- | -- | -- | -- | -- | --
BCOATA | Butyryl-CoA-ACP transacylase | acyl carrier protein[c] + Butanoyl-CoA[c] <=>   Butyryl-ACP (n-C4:0ACP)[c] + Coenzyme A[c] | 2.3.1.180 | XNR_4511 or XNR_4511 or XNR_0396 or XNR_4511 | -1000 | 1000
BCOATA_scoGEM | Butyryl-CoA-ACP transacylase | acyl carrier protein[c] + Butanoyl-CoA[c] <=> Coenzyme   A[c] | 2.3.1.180 | (SCO1271 and SCO2389) or (SCO1271 and SCO0549) or SCO1271 or   SCO1267 or (SCO1271 and SCO1272) or (SCO2388 and SCO2389) or (SCO2388 and   SCO0549) or SCO2388 or SCO1267 or (SCO2388 and SCO1272) or (SCO6564 and   SCO2389) or (SCO6564 and SCO0549) or SCO6564 or SCO1267 or (SCO6564 and   SCO1272) | -1000 | 1000

ckitti:
**Update**: Because this issue with misrepresented S matrix looked familiar as with what I was trying to address in #223 , I used the modified `mergeModels` function from branch `fix: mergeModels` and this problem no longer occurs.

PR #231, opened 2019-08-12
RAVEN 2.2.2

edkerk:
* Features:
  * New function `setMetExchange` allows users to set the exchange flux bounds for a given list of metabolites. This should be helpful for specifying e.g., media conditions, or other nutrient uptake/export fluxes. (PR #222)
  * `setParam` can set upper and lower bounds with a defined variance in percentage: setting this value to `5` results in the lower bound being set at 97.5% of the provided value, and the upper bound at 102.5%. This is inverted for negative bounds. (PR #225)
  *  `addTransport` has list of added reaction IDs as alternative. (PR #226)

* Fixes:
  * `getRxnsFromKEGG`: increased the number for preallocated reactions from 11k to 15k, as KEGG now includes reactions. (PR #229)
  * `getKEGGModelForOrganism`: fixed several small typos. (PR #229)
  * Slightly decrease tolerance when using gurobi solver, particularly when solving MILP, because `fillGaps` does not function well with less strict tolerance. (PR #226)
  - `addTransport` now allows adding multiple rxns and provides the list of added rxn IDs as optional output. (PR #226)
  - Error message in `addRxnsGenesMets`. (PR #226)
  - Fix `getModelFromHomology` bugs: keeping rxns without homologous genes while using multiple template models, and partial matches of gene ids when updating grRules: YAL001W is not YAL001W-A. (PR #226)
  - Fix `fillGaps` bug to keep all genes from template models, not only the ones annotated to gap-filling reactions. (PR #226)
  - Enable illegal character removal for grRules in `exportModel`. (PR #226)
  - Allow multiple compartments either in rows or columns in `addMets`. (PR #226)

* Chore:
  * Remove HMR2 (Human Metabolic Reactions 2), which is upgraded to Human-GEM and distributed through its own [repo](https://github.com/SysBioChalmers/Human-GEM). No functions mentioned HMR2, removal of binary is sufficient. (PR #227)


edkerk:
Merge PR #230 before merging this PR.

PR #237, opened 2019-08-17
feat: fix directory names for binaries

simas232:
Having the version names included in directories for external software is irrelevant, since this requires additional manual work to fix the corresponding MATLAB functions every time once updating these external programs. To overcome this issue, the directory names for external software no longer include their versions. Instead, such information is kept in software/versions.txt file.


simas232:
Yes, I agree it is relevant to include the guidelines for external software updates. Will therefore include a small section in [Devguidelines](https://github.com/SysBioChalmers/RAVEN/wiki/DevGuidelines) soon.

Hao-Chalmers:
It is good to remove external software versions from the folder/file names and collect them into a centralized file (e.g. versions.txt) so that software upgrading wouldn't affect the code much. Though it seems weird by putting a large number of changes (a total of 181 files) into a single commit! This probably should be avoided later.

simas232:
@Hao-Chalmers, I agree that including that many files in a single commit is not the best idea. However, if we are going to continue having RAVEN bundled with external software, the upcoming updates for the binary files would be that heavy by the numbers of the changed files like it is in this PR. For instance, an upcoming update for CD-HIT, MAFFT and HMMER would involve more than 200 changed files. The main culprit here is MAFFT, which brings 200 new files in total.

MAFFT would mean 200 new files introduced.

Hao-Chalmers:
@simas232 would it okay to have a relevant issue for discussing/planning if there will be heavy updates of binary files upcoming?

Hao-Chalmers:
> For instance, an upcoming update for CD-HIT, MAFFT and HMMER would involve more than 200 changed files. The main culprit here is MAFFT, which brings 200 new files in total.

For example of this commit [accb2c3](https://github.com/SysBioChalmers/RAVEN/pull/237/commits/accb2c3128da521aae542fe5a59565423f85b331), the changes to ~200 files probably could be partitioned into several commits that distinctly modifying/renaming: a). BLAST relevant files, b). DIMOND files/folder names, c). WoLFPSORT files/foldername, d). CD-HIT foldername, e). MAFFT folder names  f). adding`versions.txt`.

simas232:
Ah, okay @Hao-Chalmers. I thought you were about to suggest to no longer bundle external software in RAVEN and have it downloadable instead.

Your suggestion makes sense, I will include the section about the binary updates in [DevGuidelines](https://github.com/SysBioChalmers/RAVEN/wiki/DevGuidelines).

PR #246, opened 2019-09-02
fix: getElementalBalance ignored order of rxns when using strings (IDs)

JonathanRob:
The `getElementalBalance` function ignores the order of reactions if they are supplied as a list of reaction IDs, resulting in an output that is not aligned with the inputs.

This same bug was fixed in #240 for inputs of reaction indexes - here I implemented a more general solution that fixes the problem regardless of input type (index numbers, reaction IDs, or logical vector).

### DIFF
--- a/core/getElementalBalance.m
+++ b/core/getElementalBalance.m
@@ -133,9 +133,10 @@ for i=1:numel(toPrint)
     end
 end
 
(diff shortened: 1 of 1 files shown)


edkerk:
@JonathanRob FYI, if you're done with the development, you can request a review on the right, so that your changes can get merged in `devel`.

PR #248, opened 2019-09-11
chore/doc/feat: update KEGG to Release 91.0 + update KEGG related binaries + other improvements

simas232:
- chore/doc: updated the external binaries, which are utilised by `getKEGGModelForOrganism`. Also added the license files for this software. The binary CD-HIT file for Windows was added for the first time.
- chore: updated the KEGG module to Release 91.0. The new KEGG Orthology specific HMM sets were also uploaded and made available for the automatic download using `getKEGGModelForOrganism`.
- feat: added the possibility to run the protein clustering with CD-HIT in Windows, thanks to the newly added CD-HIT binary. The model reconstruction module from KEGG is therefore now fully compatible with all three operating systems: macOS, Windows and Unix.
- feat: the printed messages in `getKEGGModelForOrganism` and dependable functions were completely revised and rewritten for the clarity and comprehension. This allows the user to faster familiarize with the whole reconstruction pipeline and also identify the most time consuming stages. However, this comes at the cost of the code readability due to the longer code. The printed messages now do not only report the completion of the stages, but also once they start. Regarding multi-FASTA files generation, sequence clustering and multiple alignent, HMMs training and HMMs search stages, the intermediate progress reports were also implemented for 25%, 50% and 75% completion. See 8deb535d9855968f3c0d7e317e8190427e60c7c5 for the further details.


Hao-Chalmers:
I tested this PR with annotation-based draft generation but got error message "FASTA file cannot be found". It turned that the fasta file-checking section (line320-325) was too forward, it should be moved backward, maybe between line330 and 331.

simas232:
@Hao-Chalmers, you are right, the check for multi-FASTA file was too forward. It was therefore moved slighly below, just after the printing the title line

simas232:
For some weird reason (it worked fine in my machine couple of weeks ago), `getKEGGModelForOrganism` kept complaining about multi-FASTA file, when running e.g. `getKEGGModelForOrganism('sco')`:
```
Error using cell/ismember (line 34)
Input A of class double and input B of class cell must be cell arrays of character vectors, unless one is a character vector.

Error in getKEGGModelForOrganism (line 328)
filePresent=ismember(fastaFile,{dirContent.name});
```

This issue was fixed in the one additional commit.

Hao-Chalmers:
@simas232 I also met an error when running `getMetaCycModelForOrganism` that complains about missing reference FASTA file `protseq.fsa` (it worked well some weeks back).

Now it could be confirmed that the bug was from `getDimonad` (and `getBlast`), due to the logic of checking file existence only from the current folder (commit 2cd23a6). However, the query and reference FASTA files are Not in the same folder both for `getKEGGModelForOrganism ` and `getMetaCycModelForOrganism`.

The original check with `for` loop seems make more sense.

edkerk:
@Hao-Chalmers The problem with checking for file existence in MATLAB path (the old approach) vs. in current directory (new approach), is that with the old approach is not explicitly clear which file is used (it depends on which folders are in the MATLAB path).

`getMetaCycModelForOrganism` has the location of `protseq.fsa` coded as: `fullfile(ravenPath,'metacyc','protseq.fsa')` where `ravenPath` is the full path to the `external` folder in the local RAVEN installation.

Can you please provide code that can replicate the error that you encounter, and start a new issue to report this?

Hao-Chalmers:
@edkerk the code I used is line39 at `tutorial2018`.

The `filePresent` reports errors by checking only the current folder, instead of the filesystem.
>Error using getDiamond (line 60)
>FASTA file /Users/haowa/Documents/PostDoc/Sysbio/GitHub/RAVEN/external/metacyc/protseq.fsa cannot be found in the current directory

edkerk:
@Hao-Chalmers I can replicate this. Will fix this and then push to this PR.

simas232:
@edkerk, the latest fix does not work for KEGG reconstruction, see an error below:

```
Undefined function or variable 'fastaFile'.

Error in checkFileExistence (line 32)
inCurrDir = ~any(strfind(fastaFile,'\')) && ~any(strfind(fastaFile,'/'));

Error in getKEGGModelForOrganism (line 327)
fastaFile=checkFileExistence(fastaFile);
```

simas232:
@edkerk, upon the latest fixes for `checkFileExistence`, everything works fine in KEGG and MetaCyc reconstruction in macOS.

simas232:
@Hao-Chalmers, could you please check `getKEGGModelForOrganism` with and without multi-FASTA file in your machine? Upon success, feel free to merge this PR

Issue #252, opened 2019-10-02
feat: enable the external binaries functionality check in Windows


simas232:
### Description of the issue:
When running `checkInstallation`, the functionality check for external binaries is performed in macOS and Unix systems, but not in Windows. Such functionality check should be enabled for all the systems with no exceptions.

#### Reproducing this issue:
```matlab
checkInstallation
```

#### System information
1. RAVEN version: all
2. Operating system: Windows

**I hereby confirm that I have:**
- [X] Followed the [guidelines](https://github.com/SysBioChalmers/RAVEN/wiki/Installation) to install RAVEN.
- [X] Checked that a similar issue does not [already exist](https://github.com/SysBioChalmers/RAVEN/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [ ] If suitable, needed, asked first in the [Gitter chat room](https://gitter.im/SysBioChalmers/RAVEN) about the issue

Issue #234, opened 2019-08-16
feat: checkInstallation prints correct RAVEN version 


edkerk:
### Description of the issue:
When running `checkInstallation`, the first line of output details the RAVEN version as 
```matlab
*** THE RAVEN TOOLBOX v. 2.0 ***
```
The version number is hard-coded in `checkInstallation`. Instead, it would be good to read the current version from `version.txt` which is in the root of the repository in the `master` branch.

`version.txt` does not exist in the development branches (as the code will be modified and not be identical to one of the releases). Therefore, `checkInstallation` should do the following:

1. Check for existence of `version.txt` in the repo root. If it exists, read the file content and print the number in the first line of output.
2. If `version.txt` does not exist, then `checkInstallation` should instead print `*** THE RAVEN TOOLBOX - DEVELOPMENT VERSION ***` as output

The benefit is that we don't need to hard-code the version in `checkInstallation`, it will now print the x.y.z format (currently x.y is hardcoded), and it will make obvious when the user is using a development version.

#### Reproducing this issue:
```matlab
checkInstallation
```

#### System information
1. RAVEN version: all
2. Operating system: all

**I hereby confirm that I have:**
- [x] Followed the [guidelines](https://github.com/SysBioChalmers/RAVEN/wiki/Installation) to install RAVEN.
- [x] Checked that a similar issue does not [already exist](https://github.com/SysBioChalmers/RAVEN/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [ ] If suitable, needed, asked first in the [Gitter chat room](https://gitter.im/SysBioChalmers/RAVEN) about the issue

PR #254, opened 2019-10-05
doc/feat/fix: enhance checkInstallation

simas232:
doc/fix:
- `checkInstallation`: divide external binaries into two categories and which are therefore reported separately. The essential binaries (`makeblastdb`, `blastp`, `hmmsearch`) are the ones which are used by most of the users, while the remaining ones (`cd-hit`, `mafft`, `hmmbuild`) are only used if KEGG FTP dump files are parsed in `getKEGGModelForOrganism`. The latter category is therefore not essential.
- `checkInstallation`, `checkFunctionUniqueness`: fix the printed messages to increase the clarity of the printed output

feat:
- `checkInstallation` prints correct RAVEN version (#234)
- `checkInstallation`: enable the external binaries functionality check in Windows (#252)
- `checkInstallation`: add check for `diamond`, which belongs to essential external binaries category, see above

### DIFF
--- a/installation/checkFunctionUniqueness.m
+++ b/installation/checkFunctionUniqueness.m
@@ -5,7 +5,7 @@ function checkFunctionUniqueness()
 %
 %   Usage: checkFunctionUniqueness()
 %
-%   Simonas Marcisauskas, 2019-08-28
+%   Simonas Marcisauskas, 2019-10-05
 %
 
 %Get the RAVEN path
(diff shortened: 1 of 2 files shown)


simas232:
@edkerk, I aligned all `OK`/`not OK` messages to the right by using tabs, as requested. If you feel that they should be even more to the right, feel free to add more tab symbols.

simas232:
@Hao-Chalmers, we can have this in mind for the future. For now let's proceed with the current setup. If we had considered implementing left alignment for `OK`, `Not OK` now, we would need to reformat the layout in general, particularly the section where the left indentation is used.

PR #256, opened 2019-10-13
fix: optimizeProb

JonathanRob:
Fixed usage of the `params` input in the `optimizeProb` function to avoid errors and ignoring of optional parameter specification.

The `optimizeProb` function ignores the `params` input when using the Gurobi solver, preventing users (or other calling functions) to modify optimization parameters using this input. Attempting to fix this by un-commenting line 30 (`gparams=structUpdate(gparams,params);`) results in warnings from Gurobi due to the presence of Mosek-specific parameters, as well as errors due to some functions calling `optimizeProb` with `params` set to an empty double `[]`.

The problem with the Gurobi warnings was fixed by removing non-Gurobi fields (specifically `relGap` and `MSK_OPTIMIZER_FREE_SIMPLEX`) from `params` prior to passing to the `gurobi` function. This is probably not the most efficient way to do this since it is not a very generalized solution, but it works for now.

The error caused by an empty `params` input was solved by adding a second condition to the input argument check on line 14 (i.e., `|| isempty(params)`), which converts it to an empty structure that is compatible with the rest of the function (specifically the `structUpdate` sub-function).


edkerk:
The changes from this PR have been merged with PR #258.

Issue #245, opened 2019-09-02
chore: abandon MOSEK support


edkerk:
### Description of the issue:
MOSEK is problematic:
- There have been issues with MOSEK before (e.g. #208).
- The recent fix for `fillGaps` conflicts with MOSEK ([see Gitter](https://gitter.im/SysBioChalmers/RAVEN?at=5d69df8f0d4b1f5d888b0bcf)).
- Current support is upto version 7 (while MOSEK version 9 has been released).
- Keeping MOSEK working is extra effort.

Meanwhile;
- There are excellent alternatives with Gurobi (preferred anyway), or alternatively parsing the LP via COBRA)
- [Long term goal](https://github.com/SysBioChalmers/RAVEN/projects/1) is to combine with COBRA, so keeping MOSEK working with our own code is wasting effort.

Therefore:
- MOSEK support will be abandoned from version 2.3.0. 

#### System information
1. RAVEN version 2.3.0+

**I hereby confirm that I have:**
- [x] Followed the [guidelines](https://github.com/SysBioChalmers/RAVEN/wiki/Installation) to install RAVEN.
- [x] Checked that a similar issue does not [already exist](https://github.com/SysBioChalmers/RAVEN/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [ ] If suitable, needed, asked first in the [Gitter chat room](https://gitter.im/SysBioChalmers/RAVEN) about the issue

edkerk:
When releasing v2.3.2, modify Installation instructions on Wiki to remove mentions of MOSEK.

PR #258, opened 2019-10-14
fix: remove Mosek, refactor LP and MILP parsing, optimizeProb bug

edkerk:
Addressing issue #245, LP and MILP problem structures and results are now formatted in COBRA style (was MOSEK style).

PR #256 (author: @JonathanRob) is merged with this PR.

_Simple_ LP has been tested (`solveLP` with a classic GEM), and `fillGaps` also seems to work fine, but additional test runs are encouraged.


edkerk:
@JonathanRob Can you specifically confirm that this also fixes #256? Thanks!

JonathanRob:
Hmm, it looks like `runINIT` now encounters an error. The `runINIT` function sets up the `prob` structure but it doesn't have a `b` field, so line 4 in `cobraToGurobiProb` (`gprob.rhs = prob.b;`) errors with missing field `b`. Instead, `prob` from `runINIT` has `bux`, `blx`, `buc`, and `blc` - I'm not super familiar with the different solver parameters, so I'm not sure how this should be translated. Here is the full error message if that helps:

```
Error in cobraToGurobiProb (line 4)
gprob.rhs = prob.b;

Error in optimizeProb (line 30)
res = gurobi(cobraToGurobiProb(prob),gparams);

Error in runINIT (line 308)
res=optimizeProb(prob,params);
```

edkerk:
Thanks @JonathanRob. The `bux`, `blx`, `buc` and `blc` where used in the previous MOSEK-style `prob`. I'll fix this and check with you again.

edkerk:
@simas232 @JonathanRob @Hao-Chalmers could you please check whether this works for `runINIT`?

edkerk:
I noticed that in the release notes for 2.3.0 is mentioned that PR #256 was solved, but that PR was actually closed as it's changes are also included here. So this PR should be reviewed and merged with `devel` to solve PR #256, in addition to the issues mentioned in this PR (removal of MOSEK, refactoring of LP and MILP problems).

JonathanRob:
I was able to repair tINIT functionality with the new problem structure, as well as improve some features.

The main problem was that the specification of which variables were integers in the MILP setup was not formatted correctly in the new `prob` structure. This was resolved by adding variable types to the `prob.vartype` field (d8795a8), which also required updating the `cobraToGurobiProb` (e16a2cd) and `optimizeProb` (bfd5199) functions to recognize/use this new field.

An additional error was caused by the `gurobiToCobraRes` function, which would consider the optimization failed if the `vbasis` and `cbasis` fields were missing from the solution structure. Since these fields are not available for MILPs, the function has been modified to skip this part if the problem was a MILP (5c1a0fd). The `optimizeProb` function now also informs the `gurobiToCobraRes` function whether the problem is a MILP (60e15dc).

Finally, I modified the `gurobiToCobraRes` function to consider a MILP as optimized/successful if the time limit is reached, as long as it has found a solution(s) within the time limit (6fd85c1). This is necessary for tINIT, which for large models will virtually always exceed the time limit.

After implementing these changes, I ran a test with tINIT using identical data as a previous run I had done with the old version. The run behaved similarly, and yielded models that were very close in structure (as close as one can expect with the inherent stochasticity of the algorithm) to the previous results. I think it is safe to say that tINIT is now fully functional with these new updates, but I will continue to check closely for any other bugs that I may have missed.

edkerk:
Thank you @JonathanRob for the testing and fixing. It should really be just an issue of refactoring the problem structure, so if identical results are obtained then the refactoring was succesful. I also confirmed that your changes does not break ´fillGaps´, which I've used as test case for MILP problems.

PR #261, opened 2019-10-21
doc/feat: update tutorials

simas232:
### Main improvements in this PR:
Revise the existing RAVEN tutorials. This includes:
- re-format MS Word document file and rename it to `RAVEN tutorials.docx`
- add the description headers for the tutorial scripts
- rename `tutorial2018` to `tutorial5`. In addition, this script no longer depends to `SupplementaryTables.xlsx`, since this data is now included in `iMK1208+suppInfo.mat`
- other small fixes

**NOTE:** tutorial scripts were not renamed to include the semantic filenames (e.g. rename `tutorial0` to `tutorial0_importGEMRunFBA` or `importGEMRunFBA`), because the filenames would be particularly long. The existing file names should be enough while these tutorials are appropriately referenced in `RAVEN tutorials.docx` and `RAVEN Wiki`.

**I hereby confirm that I have:**

- [X] Tested my code on my own machine
- [X] Followed the [development guidelines](https://github.com/SysBioChalmers/RAVEN/wiki/DevGuidelines).
- [X] Selected `devel` as a target branch
- [ ] If needed, asked first in the [Gitter chat room](https://gitter.im/SysBioChalmers/RAVEN) about this PR

PR #262, opened 2019-10-24
doc: re-order tutorials

simas232:
- remove the "Exercise" definition and use only "Tutorial" instead
- re-order tutorials from 1 to 6 rather than from 0 to 5


Hao-Chalmers:
Semantic naming of filenames/commit messages are essential in helping the the usage of these tutorials. Maybe here is the chance of implementation (e.g. from `tutorial6` to `tutorial6_combineModelFromKEGGMetaCyc`).

simas232:
@Hao-Chalmers, an inclusion of semantic names in tutorial script names was in question in PR #261, I left a comment regarding this issue there:
> **NOTE:** tutorial scripts were not renamed to include the semantic filenames (e.g. rename `tutorial0` to `tutorial0_importGEMRunFBA` or `importGEMRunFBA`), because the filenames would be particularly long. The existing file names should be enough while these tutorials are appropriately referenced in `RAVEN tutorials.docx` and `RAVEN Wiki`.

In addition to this comment, I see that these 6 tutorials fill into two categories:
- demonstrations, do not requiring guidance from `RAVEN tutorials.docx`. Would benefit by having semantic names
- tutorials-exercises, requiring guidance from `RAVEN tutorials.docx`. Would not benefit by having semantic names

I think that inclusion of semantic names would require an unecessary additional work. As it is now, each tutorial has its own title in `RAVEN tutorials.docx`. While I made those clear and concise as possible, I believe that even now they are too long to include in the tutorial script names. So the review would be needed there.

Or maybe we should consider splitting the tutorial into two parts?

Hao-Chalmers:
Yeah, changes can be gradually and may be better so.

Some additional thoughts to organizing tutorials in future: 
i) detach the dependency of individual tutorials to that word file, which could then be deprecated;
ii) eventually use semantic filenames for these tutorials, since this is really convenient for new users;

PR #267, opened 2019-11-11
chore: re-add HMMER (Unix) 3.2.1 binaries

simas232:
 * chore:
   * re-add HMMER (Unix) 3.2.1 binaries due to the problems with existing ones

**NOTE:** The previous binaries were obtained from `Linuxbrew`, while the current ones were obtained from `miniconda`. It is probably better to stick with `miniconda` binaries in the future as soon as the versions for required software are available.


simas232:
These newly added binaries were also picked from miniconda. Okay, will try to stick to miniconda from now on.

PR #270, opened 2020-01-30
Add new compareMultipleModels function

JonathanRob:
Addition of a new `compareMultipleModels` function.

This function compares two or more condition-specific models generated from the same reference model using high-dimensional comparisons in the reaction-space. In addition to returning a `compStruct` structure with comparison statistics, the function can (optionally) generate figures illustrating the results:
* Heatmaps showing differences in relative number of reactions present in each subsystem
* Heatmap showing the structural similarity of models (Hamming similarity of reaction absence/presence vectors)
* A 3D tSNE plot comparing model structures (reaction presence/absence)
* Binary heatmap comparing models' ability to complete metabolic tasks (if provided)


JonathanRob:
This function has been sitting on a separate branch for some time now, and I think it's now ready to be added.

Hao-Chalmers:
I tested this function by using Hart et al. models (detailed in Human1 paper), everything went well.

Comments to the generated figures: 
 * Add text label to the tSNE plot, so that the results could be informative and clear to user
 * Consider another color scheme in the structural similarity plot, now it's quite strong (too red)

JonathanRob:
Thanks for checking this, @Hao-Chalmers.

In response to your suggestions:

> Add text label to the tSNE plot, so that the results could be informative and clear to user

tSNE plots are generally only informative when there are many points, so adding text labels by default to such a plot would become quite messy and unreadable for most applications. Instead, coloring groups of points based on some categorization variable is more appropriate for such a plot, and is already implemented using the `groupVector` input.

> Consider another color scheme in the structural similarity plot, now it's quite strong (too red)

I will leave decisions regarding the default color scheme up to @danieljcook. Users can easily change the color scheme by right-clicking the colorbar or using the `colormap` function when the plot is active.

Hao-Chalmers:
@JonathanRob your response makes sense to me. And nice to have this function.

PR #299, opened 2020-06-08
fix: correct capitalization metBiGGID

edkerk:
Fix:
- Correct capitalization of `metBiGGID` field for model conversion.


simas232:
@BenjaSanchez, @edkerk, we may have a function which checks if COBRA toolbox is available in MATLAB pathlist, e.g. `isCOBRAInstalled`. It is very likely that we would need this function in more situations than this.

edkerk:
@BenjaSanchez Good questions. The file is _supposed_ to be _exactly_ the same as the one from COBRA (I missed that I manually changed `metBiGGID` when I added the file here, and the reason I didn't notice this is indeed because of your explanation).

The reason to distribute the same file with RAVEN is indeed to prevent reliance on COBRA for `ravenCobraWrapper` (one might want to get the annotations in `metBiGGID` format instead of `metMiriams`). Why do we use the exact same file with the same file name? When RAVEN is starting to function as submodule in COBRA (#184), but there is still a need to move between model formats, then we can just directly rely on the COBRA-distributed file and remove the RAVEN-distributed version.

PR #302, opened 2020-06-23
Fix/met id conversion wrapper

BenjaSanchez:

`ravenCobraWrapper` properly identifies when a COBRA model comes with non-unique metabolite ids after removing the compartment info, as it is common with BiGG models. For instance, `pydx5p[c]` will be non-unique after removing the `[c]` suffix, so then it will be converted to `pydx5p_c`. However, it fails to detect this when going from RAVEN to COBRA, and simply adds an additional suffix: in this case  `pydx5p_c` will be converted to `pydx5p_c[c]`, which is undesired. This PR fixes that by detecting if **every** metabolite from the RAVEN model comes in the `met_comp` format, and if so replacing them by `met[comp]`.

#### Behavior Before

```matlab
>> tic;model_r = ravenCobraWrapper(model);toc
   Converting COBRA structure to RAVEN..
   Elapsed time is 1.364645 seconds.
>> tic;model_c = ravenCobraWrapper(model_r);toc
   Converting RAVEN structure to COBRA..
   Elapsed time is 4.205101 seconds.
>> disp(model_c.mets{1})
   pydx5p_c[c]
```

#### Behavior After

```matlab
>> tic;model_r = ravenCobraWrapper(model);toc
   Converting COBRA structure to RAVEN..
   Elapsed time is 1.282803 seconds.
>> tic;model_c = ravenCobraWrapper(model_r);toc
   Converting RAVEN structure to COBRA..
   Elapsed time is 4.039938 seconds.
>> disp(model_c.mets{1})
   pydx5p[c]
```

### DIFF
--- a/struct_conversion/ravenCobraWrapper.m
+++ b/struct_conversion/ravenCobraWrapper.m
@@ -90,7 +90,20 @@ if isRaven
     if all(~cellfun(@isempty,regexp(model.mets,'\[[^\]]+\]$')))
         newModel.mets=model.mets;
     else
-        newModel.mets=strcat(model.mets,'[',model.comps(model.metComps),']');
+        %Check if model has compartment info as "met_c" suffix in all metabolites:
+        BiGGformat = false(size(model.mets));
+        for i=1:numel(model.comps)
+            compPos=model.metComps==i;
+            BiGGformat(compPos)=~cellfun(@isempty,regexp(model.mets(compPos),['_' model.comps{i} '$']));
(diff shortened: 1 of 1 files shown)


edkerk:
This should definitely be fixed. Do we need to use the [comp] format for COBRA models, can it not be _comp?

BenjaSanchez:
@edkerk AFAIK cobratoolbox always enforces the `met[comp]` standard.

PR #309, opened 2020-07-17
feat: preserveQuotes flag in writeYaml

BenjaSanchez:

Following issue https://github.com/SysBioChalmers/yeast-GEM/issues/238, added a `preserveQuotes` flag in `writeYaml.m` that defaults to true to ensure a yaml compatible file. Tested it in [this fork](https://github.com/BenjaSanchez/yeast-GEM/commit/10d7858f80d2b752848fbcc8db7cea347cf237b7) of yeast-GEM and works as intended: All mets/rxns are now stored within quotes, and cobrapy's `load_yaml_model` can properly load the model with no errors.

### DIFF
--- a/io/writeYaml.m
+++ b/io/writeYaml.m
@@ -4,7 +4,7 @@ function writeYaml(model,name,preserveQuotes)
 %
 %   model           a model structure
 %   name            name that the file will have
-%   preserveQuotes  if quotes should be preserved for met/rxn names
+%   preserveQuotes  if quotes should be preserved for strings
 %                   (boolean, default=true)
 %
 %   Usage: writeYaml(model,name,preserveQuotes)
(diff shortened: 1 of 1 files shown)


Hao-Chalmers:
@BenjaSanchez I would suggest to add quotes to all string type elements, in addition to mets/rxns names.

BenjaSanchez:
@Hao-Chalmers do you mean compartment names and gene names? or other fields as well? I think that strings like ids/annotation should not have these quotes, as they could be interpreted as part of the id, making them incompatible with other parsers.

Hao-Chalmers:
@BenjaSanchez yes, I mean is to add quotes to all other string type fields as well. And this works well with python yaml parser, as well as Cobrapy import.

BenjaSanchez:
@Hao-Chalmers could you enumerate which fields do you mean by "all string type fields"? Most fields in the model structure are strings... I made this branch to resolve the issue outlined in https://github.com/SysBioChalmers/yeast-GEM/issues/238, so to not add too many unnecessary quotes (increasing the size of the yml file) IMO we should only add them where the string could realistically start with a `[` character (that's why I didn't include them in genes/compartments).

Hao-Chalmers:
@BenjaSanchez yes, "all string type fields" cover most fields in a model, and this will increase the file size a bit.

But this fix will preclude any unexpected strange characters in any string elements in a model, and works well with existing `yaml` parser.

BenjaSanchez:
@Hao-Chalmers I implemented your suggestion (locally) and did some more testing with the yeast model: You are right that the model still gets properly loaded with cobrapy. And although indeed the model increases in size, overall it's not that much (142 KB Vs 13 KB if we only change met/rxn names):

|Option|Model size|
|:-:|:-:|
|No quotes (model in `devel`)|[2,657 KB](https://github.com/BenjaSanchez/yeast-GEM/blob/4a8960d2dfba90b236e3c01c1c95f14aac6e7962/ModelFiles/yml/yeastGEM.yml)|
|Quotes only for rxn/met names|[2,670 KB](https://github.com/BenjaSanchez/yeast-GEM/blob/10d7858f80d2b752848fbcc8db7cea347cf237b7/ModelFiles/yml/yeastGEM.yml)|
|Quotes for all string types|[2,799 KB](https://github.com/BenjaSanchez/yeast-GEM/blob/d3eaf9670c239d705bbadfefb2cf5d7e32bab998/ModelFiles/yml/yeastGEM.yml)|

Finally, in terms of visualization (which is the main goal of the `.yml` file), it looks slightly more crowded:

image

TBH I don't have a strong preference towards either option, so we can go with @Hao-Chalmers option if we are willing to sacrifice a bit in visualization @simas232 @edkerk ??

simas232:
An option provided by @Hao-Chalmers seems to be the most suitable. An increase in file size is only ca 5% and can be neglected in favour of GEM compatibility. It is true that such YAML files look more crowded, but I think the strength of YAML is comparability, not a structure itself.

Hao-Chalmers:
I think this is ok to merge and proceed to resolving https://github.com/SysBioChalmers/yeast-GEM/issues/238.

Regarding the merging of `writeYaml` and `writeHumanYaml`, I'll start an issue to address the introduction of new fields required by `MetabolicAtlas`.

BenjaSanchez:
@Hao-Chalmers sorry, had forgotten that this option should only be triggered when `preserveQuotes=True`. I have fixed it now in the latest commit. Let me know if it looks good.

PR #313, opened 2020-09-08
chore: update COBRA structure fields

BenjaSanchez:

After the merge of https://github.com/opencobra/cobratoolbox/pull/1604, updated `COBRA_structure_fields.csv` to correspond to the latest version in COBRA's development branch. Also, removed the redundancy in `cobraNamespaces.csv`


BenjaSanchez:
Update: I spotted a bug when testing this on the yeast model, as when 2 KEGG pathways are annotated for 1 reaction, an undesired space is introduced in the RAVEN field, that COBRA toolbox adds by default when 2 annotations are available for a single field. From the loaded COBRA model:
```matlab
>> model.rxnKEGGPathways{1}

ans =

    'hsa00620; sce00234'
```

This space later pops up in [the `.yml` file](https://github.com/BenjaSanchez/yeast-GEM/commit/1423d79fddf2a90e123340e08aa777a2b3c2183f). I fixed this by [removing said spaces](https://github.com/SysBioChalmers/RAVEN/pull/313/commits/02c9a42c3073e85b4ce6e326283358b781eeff0f) in `ravenCobraWrapper.m`, and now the `.yml` file gets [properly saved](https://github.com/BenjaSanchez/yeast-GEM/commit/7287193775d018254bcea31d37f18bf54b49763b). @edkerk not sure if my fix is the best way to go around it, let me know what you think.

edkerk:
@BenjaSanchez The COBRA fields work, but the bug you identified with `ravenCobraWrapper` actually goes much deeper. In `xxxMiriams`, each annotation has its own row. So multiple entries of KEGG pathways all have their own mention. This simplifies export to SBML, as there they are also stored on individual lines as valid identifiers.org links. COBRA's `rxnKEGGPathways` etc. rather list them all in string, separated by '; ' as you identified. `ravenCobraWrapper` was properly dealing with this RAVEN -> COBRA, but not the other way around. So these `; ` should never have been in the RAVEN structure. Please check whether my change works also on your test code.

BenjaSanchez:
@edkerk tested it [here](https://github.com/BenjaSanchez/yeast-GEM/commit/5a34da0d0876ce8ed25df4604011a3e99668f1bf) and works well: COBRA model doesn't change and RAVEN model has the new fields in separate lines. Thanks for the fix!

edkerk (approved):

PR #320, opened 2020-12-03
Fix/remove genes

IVANDOMENZAIN:
- fix bug where `canRxnCarryFlux` (embeded in `removeGenes`) failed to evaluate if rxns with non-standardized grRules can carry any flux. These reactions are identified  first by `standardizeGrRules`, for each of them the genes to be removed are searched in the grRule and if a match is found then the rxn is bypassed and a warning message indicating the reaction ID and number is returned.

- Speed-up `removeGenes` function by restricting loops to `grRules` related to each of the genes to be removed, instead  of looping through all `grRules` for each gene case.

### DIFF
--- a/core/removeGenes.m
+++ b/core/removeGenes.m
@@ -20,57 +20,59 @@ function reducedModel = removeGenes(model,genesToRemove,removeUnusedMets,removeB
 if nargin<3
     removeUnusedMets = false;
 end
-
 if nargin<4
     removeBlockedRxns = false;
 end
-
 if nargin<5
     standardizeRules = true;
 end
-
 %Format grRules and rxnGeneMatrix:
 if standardizeRules
-    [grRules,rxnGeneMat] = standardizeGrRules(model,true);
-    model.grRules = grRules;
+    [grRules,rxnGeneMat,toCheck] = standardizeGrRules(model,true);
+    model.grRules    = grRules;
     model.rxnGeneMat = rxnGeneMat;
 end
-
 reducedModel = model;
-
 %Only remove genes that are actually in the model
 try
     ischar(genesToRemove{1});
     genesToRemove=genesToRemove(ismember(genesToRemove,model.genes));
 end
-
 if ~isempty(genesToRemove)
     indexesToRemove = getIndexes(model,genesToRemove,'genes');
     if ~isempty(indexesToRemove)
         %Make 0 corresponding columns from rxnGeneMat:
-        reducedModel.rxnGeneMat(:,indexesToRemove) = 0;
-        
-        genes        = model.genes(indexesToRemove);
-        canCarryFlux = true(size(model.rxns));
-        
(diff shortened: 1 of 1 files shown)


IVANDOMENZAIN:
`removeGenes` returned an error when trying to remove genes related to conflicting `grRules`. Genes removal from such reactions may cause unbalanced brackets in the modified `grRules`, which triggers the error when the `canRxnCarryFlux` evaluation is performed on them. This issue was found when trying to generate a context-specific **ecModel** by mapping [HOP92-GEM](https://github.com/SysBioChalmers/EnzymeConstrained_humanModels/blob/master/models/HOP92/HOP92.mat) to [ecHumanGEM](https://github.com/SysBioChalmers/ecModels/blob/a2cf0d30febaed7767ce0cb2ff6f716119ff382d/ecHumanGEM/model/ecHumanGEM_batch.mat).

IVANDOMENZAIN:
The speed-up refactoring reduced the computation time, for the aforementioned case, from 1717 to 1079 seconds (37%). 
Screen Shot 2020-12-03 at 15 16 26 
Screen Shot 2020-12-03 at 17 46 35

simas232:
@IVANDOMENZAIN, great implementation! Could you please share your script which you used to generate profile summary tables? If it is possible to have it as a separate function in RAVEN, we can include it in the `testing` directory, perhaps even in this PR.

IVANDOMENZAIN:
@simas232 this is actually a built-in functionality in MATLAB, the only thing that is needed for obtaining the profile summary is to run the function to test from a script, not from the command window, using the button `Run and time`
matlab

simas232:
@IVANDOMENZAIN, thank you for sharing this. We should use this function when suggesting tweaks for shorter runtimes in the future, although for the most simple cases it can be done with `tic` and `toc`.

JonathanRob (approved):
Ok, as you explained, it seems that the initial removal of genes using this function could result in imbalanced parentheses for non-standard grRules, which is of course not good. It looks like the previous version didn't actually error until the function was run again on a model whose grRule(s) had been "damaged" (imbalanced parentheses), so it was silently causing this problem - also not good.

But it looks like your fix should now solve the problem, and runs faster as you show, which is an added bonus. Although it would be nice to adapt the function to operate on non-standard rules, I think that's unnecessary at this point.

P.S. I notice you don't like empty lines in your code. Space can be your friend, @IVANDOMENZAIN :P

IVANDOMENZAIN:
Thanks for your review @JonathanRob. I agree, it would be good to find a solution for the **non-standard grRules**, by now they are bypassed by both `removeGenes` and `standardizeGrRules` but returning warning messages so that the user is aware of their presence. This should probably be raised as a separate issue so that more people can join the discussion.

P.S. You're right, don't know what's up with my aversion to empty lines, but I guess that they're not evil per se :P

PR #323, opened 2021-01-11
refactor: further efficiency increase of removeGenes

JonathanRob:
I was inspired by the recent nice work of @IVANDOMENZAIN to further speed up the `removeGenes` function.

The subfunction `canRxnCarryFlux` involved a loop through all model genes, which becomes very slow when checking a model with a lot of reactions and genes. I revised the function to take as input the genes that are present in the grRule being evaluated, so the function only needs to loop through that limited set.

I tested the change with ecHumanGEM to ensure that it returned identical results, and it also decreased the computation time by about 20-fold (690 sec to 36 sec).

### DIFF
--- a/core/removeGenes.m
+++ b/core/removeGenes.m
@@ -67,9 +67,10 @@ if ~isempty(genesToRemove)
                 for j = 1:numel(geneRxns)
                     index  = geneRxns(j);
                     grRule = reducedModel.grRules{index};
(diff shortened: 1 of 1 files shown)


IVANDOMENZAIN:
A very simple and nice solution for this function, I have tested its functionality and it even has helped to improve performance for [utility functions in GECKO](https://github.com/SysBioChalmers/GECKO/pull/120).

IVANDOMENZAIN (approved):

PR #347, opened 2021-06-15
fix: ravenCobraWrapper PubMed ID parsing

edkerk:
- fix
  - `ravenCobraWrapper` parsing `rxnReferences`-field and `pubmed` entries in `rxnMiriams`-field. When converting COBRA -> RAVEN, if `rxnReferences` entry is all numeric, it is assumed to be a PubMed ID and kept in `rxnMiriams`, otherwise `rxnReferences` is kept. For RAVEN -> COBRA, `rxnReferences` and `pubmed` entries from `rxnMiriams` are concatenated. Also use `pubmed` instead of `pmid` as indicating for PubMed IDs. This behaviour is now the same as if the model was directly loaded via RAVEN or COBRA.
  - `writeYaml` avoid empty entries.

### DIFF
--- a/io/writeYaml.m
+++ b/io/writeYaml.m
@@ -247,7 +247,7 @@ if isfield(model,fieldName)
         %All other fields:
         if strcmp(type,'txt')
             value = field{pos};
-            if preserveQuotes
+            if preserveQuotes && ~isempty(value)
                 value = strcat('"',value,'"');
             end
         elseif strcmp(type,'num')
--- a/io/writeYaml.m
+++ b/io/writeYaml.m
@@ -210,19 +210,19 @@ if isfield(model,fieldName)
         elseif strcmp(fieldName,'newMetMiriams')
             index = str2double(regexprep(name,'^.+_',''));
             name  = regexprep(name,'_\d+$','');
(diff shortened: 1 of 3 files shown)


Hao-Chalmers:
I have a concern regarding the delimiter used for separating multiple values in an element.

The current way is using semicolon and a space (`; `). It would be better to use only semicolon (`;`), without tailing space. This practice is simpler, easy in implementation and parsing.

@edkerk @simas232 what do you think?

simas232:
Nice fix! @Hao-Chalmers, it depends whether COBRA uses only semicolon or semicolon and a space to separate multiple values. I have not checked it yet, but will have a look during the following days.

edkerk:
COBRA indeed uses `; ` with a space and not just `;`. Another good reason is that DOIs, which could be a `rxnMiriam` entry, could have ';' as part of the entry, for instance http://doi.org/10.1002/(SICI)1097-0320(19990701)36:3%3C200::AID-CYTO9%3E3.0.CO;2-T. That would make it hard to separate entries by just `;`, while semicolon+space will never be in a DOI or _any_ identifier.

simas232 (approved):

Issue #350, opened 2021-06-17
bug: setParam multiple objective functions


anoop1988:
Hi,
I am using RAVEN tool kit for GEM reconstruction based gene expression data. But I am getting error in getINITModel2. Given below the code. Could someone help me to find the probolem,

params.MSK_DPAR_OPTIMIZER_MAX_TIME = 30000; % added for Mosek 7
params.MSK_DPAR_MIO_TOL_REL_GAP    = 0.02;  % added for Mosek 7

addpath(genpath('Human-GEM/'))
addpath(genpath('RAVEN-master/'))
addpath(genpath('mosek/'))
 
initCobraToolbox(false) 
changeCobraSolver ('gurobi', 'all');
%% Import the ref model
model=load('Human-GEM.mat'); 
model= addBoundaryMets(model.ihuman);
model=setParam(model, 'obj', 'HMR_3964', 1);
essentialTasks = parseTaskList('metabolicTasks_Essential.xlsx');
checkTasks(model, [], true, false, false, essentialTasks)
%% read data
New_exp= readtable('TPM.txt', 'Format','%s %f');
New.tissues={'pbmc'};
New.genes=New_exp.ID;
New.levels=table2array(New_exp(:,'TPM'));
New.threshold=1;

newGEM=getINITModel2(model, 'pbmc', [], [], New, [], true, [], true, true, essentialTasks, [], []);

Error -
Error in setParam (line 94)
            model.c(indexes(j))=params(j);

Error in getEssentialRxns (line 39)
        sol=solveLP(setParam(model,'obj',rxnsToCheck,-1));

Error in checkTasks (line 221)
                [~, taskEssential]=getEssentialRxns(tModel);

Error in getINITModel2 (line 302)
        [taskReport, essentialRxnMat]=checkTasks(cModel,[],printReport,true,true,taskStructure);

Error in Cmd (line 36)
newGEM=getINITModel2(model, 'pbmc', [], [], New, [], true, [], true, true, essentialTasks, [], []);

simas232:
@anoop1988, RAVEN does not have a function called `getINITModel2`. Could you please give more information e.g. how this function is different than `getINITModel`? Also, it would be great if you share the output you get when running `checkInstallation()`.

JonathanRob:
@simas232 `getINITModel2` is from the `Human-GEM` repo.

[...]

PR #352, opened 2021-06-21
fix: setParam multiple obj parameters

edkerk:
- fix:
  - `setParam` when multiple 'obj' values are changed at the same time (solves #350)


edkerk:
The change was already suggested in #350, and there it was reported to work. It also works for me in other scenarios.

JonathanRob (approved):
Looks good!

Issue #353, opened 2021-07-06
refactor: remove prefix 'M_' and 'R_' in exprotModel/importModel functions


haowang-bioinfo:
### Description of the issue:
* Prefixes `M_` and `R_` are added by `exprotModel` function while exporting, and removed by `importModel` while importing. This was initially designed to [follow](https://github.com/SysBioChalmers/RAVEN/blob/17df2147f604951705bdb2fbfcf12dac913a491e/io/importModel.m#L71) the "yeast consensus network model", which however had been replaced by `yeast-GEM` that is no longer exactly using these prefixes.
* Here it is proposed to remove these prefixes from both functions so that the original identifiers in a model can be presented in exported SBML file and avoid confusion.
* In addition, these prefixes are [not encouraged](https://github.com/sbmlteam/libsbml/issues/151) by libSBML scheme.

**I hereby confirm that I have:**
- [X] Followed the [guidelines](https://github.com/SysBioChalmers/RAVEN/wiki/Installation) to install RAVEN.
- [X] Checked that a similar issue does not [already exist](https://github.com/SysBioChalmers/RAVEN/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [ ] If suitable, needed, asked first in the [Gitter chat room](https://gitter.im/SysBioChalmers/RAVEN) about the issue

edkerk:
In principle I support this, but please note the following:

1. While documentation in `importModel` is mentioning the "yeast consensus network model", in essence it is about compatibility with COBRA toolbox. These prefixes used to be part of their "model format" (AFAIK), but they [still use it](https://github.com/opencobra/cobratoolbox/blob/25dbea46f08a9aeebcc24e0253293965d88e4ac6/src/base/io/utilities/writeSBML.m#L52) if an ID starts with a number.
2. With the above in mind, to keep compatibility with COBRA one might want to have the option to keep the current behaviour? So by default follow your suggestion here, but keep the option to I/O as COBRA model. What is your opinion?
3. The above code also raises the issue that identifiers that are defined in the model might not be valid SBML IDs. The [SBML specification](http://co.mbine.org/specifications/sbml.level-3.version-2.core.release-2.pdf) states the following (page 12), which basically means that an ID starts with either a letter or underscore, and the rest of the ID can be a combination of only letters, numbers and underscores, no other characters are allowed:

```
letter ::= ’a’..’z’,’A’..’Z’
digit ::= ’0’..’9’
[...]

PR #356, opened 2021-08-09
fix: disable adding prefix 'M_' and 'R_' in exprotModel function

Hao-Chalmers:
* This PR modify `exportModel` function to stop adding by prefix `M_` and `R_` to metabolite and reactions ids, as proposed in #353

### DIFF
--- a/io/exportModel.m
+++ b/io/exportModel.m
@@ -281,13 +281,13 @@ for i=1:numel(model.mets)
     end
     
     if isfield(modelSBML.species,'metaid')
-        modelSBML.species(i).metaid=['M_' model.mets{i}];
+        modelSBML.species(i).metaid=model.mets{i};
     end
     if isfield(modelSBML.species, 'name')
         modelSBML.species(i).name=model.metNames{i};
     end
     if isfield(modelSBML.species, 'id')
-        modelSBML.species(i).id=['M_' model.mets{i}];
+        modelSBML.species(i).id=model.mets{i};
     end
     if isfield(modelSBML.species, 'compartment')
         modelSBML.species(i).compartment=model.comps{model.metComps(i)};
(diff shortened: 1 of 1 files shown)


edkerk:
To my opinion, `importModel` should then also be changed, see my comment in #353.

edkerk:
Also, I suggest this would be part of a minor release (2.5.2 -> 2.6.0). It will break compatibility.

mihai-sysbio (changes requested):
Very nice to have this fix in Raven. I noted a few things
- [x] in `io/exportModel.m:510` this should be updated `rdf:about="#meta_R_' model.rxns{i} '"`
- [ ] there's a reference to `meta_M` in `doc/io/exportModel.html:391` and to `meta_R` in the same file on line `569`
- [x] ~I think @edkerk noted that there is no function that reports all SBML-invalid identifiers, which I agree should be part of this PR that solves #353~ see below

mihai-sysbio:
>     * [ ]  I think @edkerk noted that there is no function that reports all SBML-invalid identifiers, which I agree should be part of this PR that solves [refactor: remove prefix 'M_' and 'R_' in exprotModel/importModel functions #353](https://github.com/SysBioChalmers/RAVEN/issues/353)

On second thought, @edkerk would it be okay to resolve the concern above as a different PR, to keep PRs small and targeted? This PR will then resolve only a part of #353, which makes sense to me.

edkerk (changes requested):
- [x] Remove line 225-231, where `c_` is appended to `model.comps`.

Hao-Chalmers:
>  there's a reference to meta_M in doc/io/exportModel.html:391 and to meta_R in the same file on line 569

the html files will be corrected by the `m2html` later

edkerk:
We have the routine to run `updateDocumentation` just before making a new release (example #357), but it doesn't hurt to run it earlier already.

edkerk (approved):
Successful I/O of `yeast-GEM`.

mihai-sysbio (approved):

Hao-Chalmers:
Here I merge this since `tiny progress is still progress`.  The ID check function and `importModel` will be modified in other PRs.

PR #364, opened 2021-08-30
feat: sort model alphabetically by identifiers

edkerk:
- feat:
  - ~~`exportForGit` sorts the `model.rxns`, `model.mets` and `model.genes` fields alphabetically (plus other relevant fields from the [RAVEN model structure](https://github.com/SysBioChalmers/RAVEN/wiki/RAVEN-Model-Structure)) before writing any model output files.~~
  - `sortIdentifiers` can sort metabolites, reactions, genes and compartments alphabetically by their identifier, and is set as option (`sortIds`, default = _false_) when exporting models.

Git-based tracking of changes is much simplified if the file entries are consistently ordered, so that just swapping location in the file would not be marked as a change. Note that if the user has defined any other fields that refer to the list of reactions, metabolites or genes and is not mentioned in the definition of the [RAVEN model structure](https://github.com/SysBioChalmers/RAVEN/wiki/RAVEN-Model-Structure) it would not be ordered! If such common fields exist, they can be suggested to be part of the definition of RAVEN model structure (as optional field).


mihai-sysbio (dismissed):
Such a nice addition!

I am still quite new to understanding the exporting logic, so please bear with me: the way I look at this, it seems `exportForGit` is a "parent" function that [relies on specific functions to create the different file formats](https://github.com/SysBioChalmers/RAVEN/blob/devel/io/exportForGit.m#L78-L115). So, while it makes sense to sort the model only once in `exportForGit`, it still allows one to call the different export functions separately, in which case the model would not be sorted. This makes for a somewhat inconsistent behaviour.

One work-around that I can imagine is for all the export functions to have yet another boolean parameter (`shouldSort`?) that defaults to _true._ When calling these functions from `exportForGit`, the value would be set to _false_, so as to avoid sorting unnecessarily.

All of this being said, I see the scope of the PR is to sort _for git_. To the best of my knowledge, however, no file format relies on a specific order, so having a generic behaviour of always sorting is simpler to me, and should not affect anything.

edkerk:
@mihai-sysbio I see your argument, it could indeed be included in each of the export functions and set to true as default. A `sortIdentifiers` function could contain this code and be called by each individual export function.

edkerk:
This PR was approved earlier, but the code has changed since, so I would like it to be reviewed again, thanks!

edkerk:
@simas232 Now you have experience with running tests, could you please run `results = runtests('sortIdentifiers_and_permuteModelTests.m')` (located in testing\unit_tests\) and review this PR?

simas232 (approved):
The unit testing function works great. I just committed some minor fixes concerning variable names, comments and unnecessary load of tINIT tasks file

Issue #366, opened 2021-09-03
unintended prefix for compartment i


mihai-sysbio:
### Description of the issue:
When exporting Human-GEM from `yml` to `xml`, the compartment `i` becomes `c_i`. It is the only compartment that gets prefixed with `c_`. The root cause is likely in the lines https://github.com/SysBioChalmers/RAVEN/blob/800362e71b9d04ae2e2a16960b9d19aff346d37e/io/exportModel.m#L224-L231
Maybe due to the unintended interpretation of i as an [imaginary unit](https://en.wikipedia.org/wiki/Imaginary_unit)?

#### Reproducing this issue:
Below is a test case that demonstrates the bug:
```matlab
compartment = 'i';
if ~isnan(str2double(compartment))
compartment=strcat('c_',compartment);
end
disp(compartment)
```
```
c_i
```

#### System information
The bug is based on the latest release of RAVEN, demonstrated above in Matlab 2019b on macOS.

**I hereby confirm that I have:**
- [ ] Followed the [guidelines](https://github.com/SysBioChalmers/RAVEN/wiki/Installation) to install RAVEN.
- [x] Checked that a similar issue does not [already exist](https://github.com/SysBioChalmers/RAVEN/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [ ] If suitable, needed, asked first in the [Gitter chat room](https://gitter.im/SysBioChalmers/RAVEN) about the issue

edkerk:
You're right. Checking whether something is numeric using `str2double` is very sloppy, this is fixed by changing line 225 as:

```matlab
        if any(regexp(model.comps(i),'^\d'))
```

Will make a PR for this asap. Meanwhile, `exportModel` from #360, together with the extended `checkModelStruct` from #353 would prevent these issues.

mihai-sysbio:
> Will make a PR for this asap. Meanwhile, `exportModel` from #360, together with the extended `checkModelStruct` from #353 would prevent these issues.

Thanks @edkerk for the quick action.

PR #367, opened 2021-09-03
fix: exportModel correct prefix compartment IDs

edkerk:
- fix:
  - `exportModel` correct handling if a compartment ID is `i` (solves #366)

### DIFF
--- a/io/exportModel.m
+++ b/io/exportModel.m
@@ -222,7 +222,7 @@ for i=1:numel(model.comps)
     end
     
     if isfield(modelSBML.compartment,'metaid')
-        if any(regexp(model.comps(i),'^\d'))
+        if regexp(model.comps(i),'^[^a-zA-Z_]')
             EM='The compartment IDs are in numeric format. For the compliance with SBML specifications, compartment IDs will be preceded with "c_" string';
             dispEM(EM,false);
             model.comps(i)=strcat('c_',model.comps(i));
--- a/io/exportModel.m
+++ b/io/exportModel.m
@@ -222,7 +222,7 @@ for i=1:numel(model.comps)
     end
     
     if isfield(modelSBML.compartment,'metaid')
-        if ~isnan(str2double(model.comps(i)))
+        if any(regexp(model.comps(i),'^\d'))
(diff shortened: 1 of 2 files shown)


io/exportModel.m
```diff
@@ -222,7 +222,7 @@ function exportModel(model,fileName,exportGeneComplexes,supressWarnings)
     end
     
     if isfield(modelSBML.compartment,'metaid')
-        if ~isnan(str2double(model.comps(i)))
+        if any(regexp(model.comps(i),'^\d'))
```
mihai-sysbio, line 225: Would it make sense to also avoid other non-letter characters, or is that taken care of at some other point?

```suggestion
        if any(regexp(model.comps(i),'^[^a-zA-Z]'))
```

edkerk: You're right that `^\d` is actually too narrow, but besides letters also underscore is [allowed](https://github.com/SysBioChalmers/RAVEN/issues/353#issuecomment-875004745) at beginning of `SId` in SBML. Would that make `^[a-zA-Z_]` as regex?

mihai-sysbio: The first `^` is an anchor for the beginning of string, and the second `^` is for the negative pattern. It is of course possible to remove the negative pattern as suggested above, but then `any` should turn into `~any`. With the suggestion above, a compartment like `no_prefix_be4_no` would be end up prefixed.

Below is the updated suggestion including the accepted `_`, and removing the `any` which is redundant.
```suggestion
        if regexp(model.comps(i),'^[^a-zA-Z_]')
```

edkerk: You're of course correct :), confusing that `^` can have multiple meanings (anchor and negation)!

mihai-sysbio: Hehe I had to look it up myself to be sure. I saw somebody use https://regex101.com and now that's my go to when in doubt of regex patterns.
Thanks for looking up the SBML pattern, it would have been so easy to miss that `_`.

mihai-sysbio (approved):
The code looks good to me, thanks for the discussion @edkerk.
I'm not sure if the documentation should also be updated before merging.

PR #369, opened 2021-09-12
feat: speed up getKEGGModelForOrganism

edkerk:
- refactor:
  - speed up `getKEGGModelForOrganism` when renaming gene names from KEGG.
- feat:
  - allow `getKEGGModelForOrganism` to take a `globalModel` structure containing `model` and `KOModel` output from `getModelFromKEGG` as preload, to speed up generating many models for different species.


edkerk:
When running `getKEGGModelForOrganism` many times for different species, it would speed up to take a preloaded global KEGG model as input, instead of running `getModelFromKEGG` with each iteration. Example of its use in tutorial6:

```matlab
[globalModel.model,globalModel.KOModel] = getModelFromKEGG;
ScoKEGGHomology=getKEGGModelForOrganism('ScoKEGGHMMs','Sco_all_protein.faa','prok90_kegg94','',1,0,0,0,10^-50,0.3,0.8,inf,inf,-1,globalModel);
```

Note that all other parameter values now also need to be given as input (and here the default values are given). It is not the most convenient, but left this way as it will not be run as such very often. Regardless, it should speed up running the function, as `getModelFromKEGG` only needs to be run once for all models.

simas232 (approved):

PR #384, opened 2021-12-06
feat: automated testing

mihai-sysbio:
To simplify development, there should be a way for tests to run automatically. This PR aims to resolve this by adding basic support to run the currently defined tests.

This PR is based on a version of `devel` that has the latest from the main branch, which is why there is long list of commits. I found this to be the easiest way of pulling in the `.github` folder that was needed, which seemed to exist only on the master branch.

- [ ] Tested my code on my own machine

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,12 +1,12 @@
 ### Description of the issue:
-*Try to be as clear as possible, e.g.:
+<!-- Try to be as clear as possible, e.g.:
 - Is there a functionality missing?
 - Did an existing function fail?
 - What is the exact error message, and other output that you got?
-- What have you tried yourself to fix it?
+- What have you tried yourself to fix it? -->
 
 #### Reproducing this issue:
-*If applicable, please attach the problematic code. PLEASE DELETE THIS LINE.*
+<!-- If applicable, please attach the problematic code. -->
 ```matlab
 INSERT CODE HERE
 ```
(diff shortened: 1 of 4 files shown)


mihai-sysbio:
Maybe it would be more useful if the test results would be added as a comment to a PR?

edkerk:
The reason why all the `updateVersion` commits are only applied to `main` and not `devel` (the reason for the long list of commits here), is that `main` contains `version.txt` which is updated with each release. `devel` does not have this file, to avoid confusion that the status of `devel` is not representative of a particular release (except just after it has been pushed to `main`).

Meanwhile, I don't see any strong reason why `.github` is only in `main`. I suggest:
- Make a separate PR to "copy" the `.github` from `main` to `devel`
  - During the next PR to `main` this might results in a conflict, but should be easily resolved
- Reapply 9f153e59111773c434dbc601db1f6d97066ac10b to a new feat-branch that is based on the updated `devel`.

These two steps can actually be combined as well.

mihai-sysbio:
> But yes, might be more _accessible_ to directly see the test results. However, what if further commits are added to the same PR, does it leave a comment with each test run?

Indeed, but we could make it conditional, i.e. only on test failures.

.github/issue_template.md
```diff
@@ -0,0 +1,24 @@
...
+INSERT CODE HERE
+```
+
+#### System information
+* Please report:
+1. RAVEN version (stabile [release](https://github.com/SysBioChalmers/RAVEN/releases), `devel` branch?)
+2. Operating system (Windows/Mac/Linux; include version)
+
+**I hereby confirm that I have:**
+<!-- Note: replace [ ] with [X] to check the box. PLEASE DELETE THIS LINE -->
```
mihai-sysbio, line 20:
```suggestion
<!-- Note: replace [ ] with [X] to check the box -->
```

edkerk (approved):

edkerk:
As you suggested above, the GitHub Action can post a comment with the MATLAB command line output?

mihai-sysbio:
@edkerk the PR commenting has been implemented in this PR.
I've been trying to split the output on multiple lines, but didn't quite get that to work. I'm hoping it's acceptable for now.
If anyone has any ideas, feel free to try them directly on this branch.

edkerk (commented):
Looks very nice! But now only the first line is shown. Perhaps better to show all results, even if it is not split over separate lines? Reverting the last commit will do this?

mihai-sysbio:
> Looks very nice! But now only the first line is shown. Perhaps better to show all results, even if it is not split over separate lines? Reverting the last commit will do this?

Yes, I removed the last commit and the PR comment was consequently updated.

edkerk (approved):

PR #411, opened 2022-04-22
fix: posting results from tests via Actions

mihai-sysbio:

This PR fixes the parsing of the tests result that are run with GH Actions, thus improving the short version of the results that gets posted automatically to the PR thread.

- [ ] Tested my code on my own machine
- [x] If needed, asked first in the [Gitter chat room](https://gitter.im/SysBioChalmers/RAVEN) about this PR

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -14,8 +14,8 @@ jobs:
         id: matlab-test
         run: |
           TEST_RESULTS=$(/usr/local/bin/matlab -nodisplay -nosplash -nodesktop -r "addpath(genpath('.')); cd('testing/unit_tests'); runtests(struct2table(dir('*.m')).name); exit;")
-          PARSED_RESULTS=$(echo $TEST_RESULTS | awk -F'com\\.' '{ n = split($2, v, "(___+)|(===+)|(---+)"); for (i = 0; ++i <= n;) { print v[i] } }')
-          echo ::set-output name=results::$PARSED_RESULTS
+          PARSED_RESULTS=$(echo "$TEST_RESULTS" | awk -F'com\\.' '{ n = split($2, v, "(___+)|(===+)|(---+)"); for (i = 0; ++i <= n;) { print v[i] } }')
+          echo ::set-output name=results::"$PARSED_RESULTS"
 
       - name: Post comment
         uses: NejcZdovc/comment-pr@v1
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -14,7 +14,7 @@ jobs:
         id: matlab-test
         run: |
           TEST_RESULTS=$(/usr/local/bin/matlab -nodisplay -nosplash -nodesktop -r "addpath(genpath('.')); cd('testing/unit_tests'); runtests(struct2table(dir('*.m')).name); exit;")
-          PARSED_RESULTS=$(echo $TEST_RESULTS | awk -F'com\.' '{ n = split($2, v, "(___+)|(===+)|(---+)"); for (i = 0; ++i <= n;) { print v[i] } }')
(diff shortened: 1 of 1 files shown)


edkerk:
It seems like the MATLAB output is still cut after a certain length? Looking at the edit history of the github-actions bot [post](https://github.com/SysBioChalmers/RAVEN/pull/411#issuecomment-1106073698), it seems like the post from 22 April 2022 10:11 CEST showed the longer text?

Also, I'm not sure how this exactly works, but would it be possible to keep the line-breaks in the MATLAB output? Considering it is a markdown document, maybe each sentence should end with two spaces?

.github/workflows/testing.yml
```diff
@@ -7,22 +7,21 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
 
-    - name: Checkout
-      uses: actions/checkout@v2
+      - name: Run tests
+        id: matlab-test
+        run: |
+          TEST_RESULTS=$(/usr/local/bin/matlab -nodisplay -nosplash -nodesktop -r "addpath(genpath('.')); cd('testing/unit_tests'); runtests(struct2table(dir('*.m')).name); exit;")
+          PARSED_RESULTS=$(echo $TEST_RESULTS | awk -F'com\.' '{ n = split($2, v, "(___+)|(===+)|(---+)"); for (i = 0; ++i <= n;) { print v[i] } }')
```
mihai-sysbio, line 17: This is the core part of the fix:
```suggestion
          PARSED_RESULTS=$(echo $TEST_RESULTS | awk -F'com\\.' '{ n = split($2, v, "(___+)|(===+)|(---+)"); for (i = 0; ++i <= n;) { print v[i] } }')
```

mihai-sysbio:
> It seems like the MATLAB output is still cut after a certain length? Looking at the edit history of the github-actions bot [post](https://github.com/SysBioChalmers/RAVEN/pull/411#issuecomment-1106073698), it seems like the post from 22 April 2022 10:11 CEST showed the longer text?

Right! This is resolved in 867189a79d869c0f5720fc03f844ae91e8e7ab51.

> Also, I'm not sure how this exactly works, but would it be possible to keep the line-breaks in the MATLAB output? Considering it is a markdown document, maybe each sentence should end with two spaces?

I've tried, but the text is shared via a bash variable and at some point the newline gets removed.

.github/workflows/testing.yml
```diff
@@ -7,22 +7,21 @@ jobs:
...
+        uses: actions/checkout@v2
 
-    - name: Checkout
-      uses: actions/checkout@v2
+      - name: Run tests
+        id: matlab-test
+        run: |
+          TEST_RESULTS=$(/usr/local/bin/matlab -nodisplay -nosplash -nodesktop -r "addpath(genpath('.')); cd('testing/unit_tests'); runtests(struct2table(dir('*.m')).name); exit;")
+          PARSED_RESULTS=$(echo $TEST_RESULTS | awk -F'com\\.' '{ n = split($2, v, "(___+)|(===+)|(---+)"); for (i = 0; ++i <= n;) { print v[i] } }')
+          echo ::set-output name=results::$PARSED_RESULTS
```
edkerk, line 18:
```suggestion
          PARSED_RESULTS=$(echo "$TEST_RESULTS" | awk -F'com\\.' '{ n = split($2, v, "(___+)|(===+)|(---+)"); for (i = 0; ++i <= n;) { print v[i] } }')
          echo ::set-output name=results::"$PARSED_RESULTS"
```

Might this resolve the issue with the line breaks, as suggested [here](https://stackoverflow.com/questions/22101778/how-to-preserve-line-breaks-when-storing-command-output-to-a-variable)?

edkerk (changes requested):

.github/workflows/testing.yml
edkerk: This should have been a "comment", not "request changes".

mihai-sysbio: Thanks for the input - it seemed the problem is how GH Actions is dealing with some chars.

Issue #414, opened 2022-05-02
feat: unit tests for functions


edkerk:
The `testing/unit_tests` folder now already includes tests for many functions (as of 5a65fe0455ac41e6cab4826981cf268a4127f057). Ideally this should cover all functions (`*.m` files, except from those in the `legacy` and `software` folders). Here the progress is tracked. Some functions do not have their own tested, but are run when another function is tested. The tests are typically not comprehensive: they do not test _all_ possible functionality (for instance different scenarios where different types of input are provided), but they are testing at least the most common use of each function.

 
<details>
  <summary>Click here to show the list</summary>

- [ ] INIT\getINITModel.m
- [ ] INIT\runINIT.m
- [ ] core\FSEOF.m
- [x] core\addExchangeRxns.m
- [x] core\addGenesRaven.m
- [x] core\addMets.m
- [x] core\addRxns.m
- [x] core\addRxnsGenesMets.m
- [x] core\addTransport.m
- [ ] core\analyzeSampling.m
- [x] core\buildEquation.m (via `addRxns`)
- [x] core\canConsume.m
- [x] core\canProduce.m
- [x] core\changeGeneAssoc.m
- [x] core\changeGrRules.m
- [x] core\changeRxns.m
- [x] core\checkModelStruct.m (via `exportModel`)
- [x] core\checkProduction.m
- [ ] core\checkRxn.m
- [ ] core\checkTasks.m
- [ ] core\compareMultipleModels.m
- [ ] core\compareRxnsGenesMetsComps.m
- [x] core\constructEquations.m (via `exportToExcelFormat`)
- [x] core\constructS.m (via `addRxns`)
- [x] core\consumeSomething.m
- [ ] core\contractModel.m
- [x] core\convertCharArray.m (via many functions)
- [ ] core\convertToIrrev.m
- [x] core\copyToComps.m
- [ ] core\deleteUnusedGenes.m
- [ ] core\dispEM.m
- [ ] core\expandModel.m
- [x] core\fillGaps.m
- [x] core\findGeneDeletions.m
- [x] core\findRAVENroot.m (via many functions)
- [ ] core\fitParameters.m
- [ ] core\fitTasks.m
- [ ] core\followChanged.m
- [ ] core\followFluxes.m
- [ ] core\gapReport.m
- [x] core\generateNewIds.m (via `addRxns`)
- [x] core\getAllRxnsFromGenes.m
- [ ] core\getAllSubGraphs.m
[...]

Issue #397, opened 2022-02-15
refactor: allow string parameters to be given as string array


edkerk:
### Description of the issue:
All string parameters for RAVEN functions should be given as character array, e.g. `'r_1234'`, and not string array, e.g. `"r_1234"`. This is not well documented, and neither would there be a reason why it has to be string array. It therefore raises strange non-intuitive errors.

Either we very explicitly mention that it has to be character array in each function documentation; or we just allow both string and character arrays. For this, each function with a string parameter (e.g. `param`) could have `param = char(param)`; in the beginning of the function.

(as raised on [Gitter](https://gitter.im/SysBioChalmers/RAVEN?at=620bf76a3349fe1c7112b11f))

#### System information
* Please report:
1. RAVEN version: 2.6.0
2. Operating system: All

**I hereby confirm that I have:**
- [X] Followed the [guidelines](https://github.com/SysBioChalmers/RAVEN/wiki/Installation) to install RAVEN.
- [x] Checked that a similar issue does not [already exist](https://github.com/SysBioChalmers/RAVEN/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [ ] If suitable, needed, asked first in the [Gitter chat room](https://gitter.im/SysBioChalmers/RAVEN) about the issue

PR #415, opened 2022-05-02
refactor: char, string and cell array input

edkerk:
- feat:
  - allow both character vectors (`'example'`) and string arrays (`"example"`) as input when text strings are expected (solves #397)
  - when a function expects a cell array as input, also allow character vector and string arrays instead of forcing a 1x1 cell array (and make the necessary conversion)
  - numerous tests (updating #414)

- fix:
  - MAFFT: explicitly force `mafft.bat` to be handled as binary, to prevent end-of-line character changes when checking out on Windows


johan-gson (approved):
Tested the test cases on my computer using
"runtests(struct2table(dir('*.m')).name)" - all but the mafft tests worked, which according to Ed was fine.

Issue #469, opened 2023-03-03
fix: addRxns definition of `rev` field


edkerk:
### Description of the issue:
https://github.com/SysBioChalmers/RAVEN/blob/6638c2bab213e143fe690e6f05d7a1e9ce953194/core/addRxns.m#L217

To define `rev`, both LB and UB should be checked. Only if lb < 0 and ub > 0 should rev = 1.

PR #473, opened 2023-03-12
fix: addRxns reverse field

edkerk:
- fix:
  - `addRxns` correctly defines `rev` field, only reactions that can go in both directions (LB<0 and UB>0) are reversible (solves #469)

### DIFF
--- a/core/addRxns.m
+++ b/core/addRxns.m
@@ -214,7 +214,7 @@ else
     for i = 1:length(rxnsToAdd.mets)
         mets         = rxnsToAdd.mets{i};
         stoichCoeffs = rxnsToAdd.stoichCoeffs{i};
-        isrev        = rxnsToAdd.lb(i) < 0;
+        isrev        = rxnsToAdd.lb(i) < 0 && rxnsToAdd.ub(i) > 0;
         rxnsToAdd.equations{i} = buildEquation(mets,stoichCoeffs,isrev);
     end
 end
--- a/core/addRxns.m
+++ b/core/addRxns.m
@@ -214,7 +214,11 @@ else
     for i = 1:length(rxnsToAdd.mets)
         mets         = rxnsToAdd.mets{i};
         stoichCoeffs = rxnsToAdd.stoichCoeffs{i};
-        isrev        = rxnsToAdd.lb(i) < 0 && rxnsToAdd.ub(i) > 0;
+        if isfield(rxnsToAdd,'ub')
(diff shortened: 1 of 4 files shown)


core/addRxns.m
```diff
@@ -214,7 +214,11 @@
     for i = 1:length(rxnsToAdd.mets)
         mets         = rxnsToAdd.mets{i};
         stoichCoeffs = rxnsToAdd.stoichCoeffs{i};
-        isrev        = rxnsToAdd.lb(i) < 0;
+        if isfield(rxnsToAdd,'ub')
+            isrev        = rxnsToAdd.lb(i) < 0 && rxnsToAdd.ub(i) > 0;
+        else
+            isrev        = rxnsToAdd.lb(i) < 0;
```
ae-tafur, line 220: a reaction can also be irreversible if lb 0 andt ub > 0. Then this line will result in a logical 0
```suggestion
            isrev        = rxnsToAdd.lb(i) == 0 ||  rxnsToAdd.ub(i) == 0;
```

core/addRxns.m
```diff
@@ -214,7 +214,11 @@
     for i = 1:length(rxnsToAdd.mets)
         mets         = rxnsToAdd.mets{i};
         stoichCoeffs = rxnsToAdd.stoichCoeffs{i};
-        isrev        = rxnsToAdd.lb(i) < 0;
+        if isfield(rxnsToAdd,'ub')
```
ae-tafur, line 217: if field `ub` does not exits, it will assume ub 1000 or inf ?

in any case if not exits, ub needs to be defined here, isn't it?

core/addRxns.m
edkerk: But note that line 220 is only run if no 'ub' is specified (`addRxns` allows for that). If no 'ub' is specified, `addRxns` will set ub to `defaultUB` or `1000`, so >0. If 'lb' is then <0, the reaction is reversible.

core/addRxns.m
edkerk: `ub` is actually defined later, and this solution here was simpler than refactoring and moving the following code _before_ the definition of reversibility.

https://github.com/SysBioChalmers/RAVEN/blob/8cff16e8824dedb41dd75ee2dc18ddd6b097d811/core/addRxns.m#L294-L315

core/addRxns.m
ae-tafur: Ok, I got it

ae-tafur (approved):
Looks good !

PR #499, opened 2023-06-30
RAVEN 2.8.3

### LINKED ISSUE #492: bug: RAVEN 2.8.2 checkInstallation on new install
### Description of the issue:
As reported on [Gitter](https://matrix.to/#/!ilVpnqUWbQRFARrPdk:gitter.im/$ylo4tZJWeROaRekxLWR5kHcm8LK7dJPNaqI1JGHwpaQ?via=gitter.im&via=matrix.org&via=matrix.freyachat.eu): 

image

This occurs if RAVEN is not yet in the MATLAB path, for instance if this is a fresh RAVEN installation.

As work-around:

> Open the file RAVEN/core/printOrange.m, copy the whole text and paste it at the end of RAVEN/checkInstallation.m (line 343), save checkInstallation and rerun it.

Code is also modified here.Solution is provided here.

#### Reproducing this issue:
In a MATLAB instance that not yet has RAVEN in its path:
```matlab
checkInstallation
```

#### System information
RAVEN 2.8.2 only.

- [x] Checked that a similar issue does not [already exist](https://github.com/SysBioChalmers/RAVEN/issues?utf8=%E2%9C%93&q=is%3Aissue)

edkerk:
- fix:
  - `checkFunctionUniqueness` on Mac/Unix with non-RAVEN paths.
  - `checkInstallation` find `printOrange` on new RAVEN installations (closes #492).
  - `readYAMLmodel` allows for a model that has gene entries that are not present in any grRules: these are unused genes and should be allowed. This also prevents a bug where gene annotations are wrongly assigned.
  - `writeYAMLmodel` correct parsing of text to avoid escape characters (e.g. SMILES may contain \C).
  - `importModel` assigns -Inf and Inf if no LB and/or UB values are defined.
  - `ravenCobraWrapper` makes empty model.c if COBRA model had no `c` field (is not necessarily made by readCbModel)

- feat:
  - `keggPhylDist.mat` includes species names.
  - `getMetaCycModelForOrganism` input and output of blastStructure (before cutoffs applied), so that function can repeatedly be run with different cutoffs without having to rerun blast/diamond.

- [x] Selected `main` as a target branch


edkerk:
@mihai-sysbio The GitHub Action shows to use RAVEN 2.8.2, but should it not use the RAVEN from this PR?

mihai-sysbio:
> @mihai-sysbio The GitHub Action shows to use RAVEN 2.8.2, but should it not use the RAVEN from this PR? `checkInstallation` should then not show a version, but rather report `DEVELOPMENT`.

According to this line:
https://github.com/SysBioChalmers/RAVEN/blob/0a180fed494b6c38a6ad9433c625898423972336/.github/workflows/tests.yml#L16

it should pick up whatever is in the current branch.

The base branch is `main`, which has a version - `2.8.2`. The branch that is merged in (`develop`) doesn't change the `version.txt` file, so RAVEN "thinks" the current code is `2.8.2`. The version is then bumped directly only `main`, after this PR, with 0a180fed494b6c38a6ad9433c625898423972336 . I've re-ran the action, and you can see in the comment that it picked up that a new version exists. Since that is "outside" of this PR, it still thinks it's running `2.8.2`.

One wait to around this would be to create a special workflow only for PRs that target `main` which remove the `version.txt` file, so the workflow would then report PRs such as this one as `DEVELOPMENT`.

edkerk:
As long as we're sure it installs RAVEN fresh from the current PR, then it's fine for me.

Issue #527, opened 2024-02-23
bug: make binaries executable on Ubuntu


edkerk:
### Discussed in https://github.com/SysBioChalmers/RAVEN/discussions/526

<div type='discussions-op-text'>

<sup>Originally posted by **juferprada97** February 21, 2024</sup>
I have the following problem when installing Raven in matlab (Liinux64 - ubuntu) with the "make binaries executables".

line 1: syntax error near unexpected token `('
/bin/bash: -c: line 1: `chmod +x
"/home/jufer/Apps/Raven/RAVEN/software/blast+/blastp".(mexa64|mexglx|mexmaci64|mac|bat)'  
> In checkInstallation>makeBinaryExecutable (line 338)
In checkInstallation (line 107) 
Warning: Failed to make /home/jufer/Apps/Raven/RAVEN/software/blast+/makeblastdb executable: /bin/bash:
-c: line 1: syntax error near unexpected token `('
/bin/bash: -c: line 1: `chmod +x
"/home/jufer/Apps/Raven/RAVEN/software/blast+/makeblastdb".(mexa64|mexglx|mexmaci64|mac|bat)'  
> In checkInstallation>makeBinaryExecutable (line 338)
In checkInstallation (line 107) 
Warning: Failed to make /home/jufer/Apps/Raven/RAVEN/software/cd-hit/cd-hit executable: /bin/bash: -c:
line 1: syntax error near unexpected token `('
/bin/bash: -c: line 1: `chmod +x
"/home/jufer/Apps/Raven/RAVEN/software/cd-hit/cd-hit".(mexa64|mexglx|mexmaci64|mac|bat)'  
> In checkInstallation>makeBinaryExecutable (line 338)
In checkInstallation (line 107) 
Warning: Failed to make /home/jufer/Apps/Raven/RAVEN/software/diamond/diamond executable: /bin/bash:
-c: line 1: syntax error near unexpected token `('
/bin/bash: -c: line 1: `chmod +x
"/home/jufer/Apps/Raven/RAVEN/software/diamond/diamond".(mexa64|mexglx|mexmaci64|mac|bat)'  
> In checkInstallation>makeBinaryExecutable (line 338)
In checkInstallation (line 107) 
Warning: Failed to make /home/jufer/Apps/Raven/RAVEN/software/hmmer/hmmbuild executable: /bin/bash: -c:
line 1: syntax error near unexpected token `('
/bin/bash: -c: line 1: `chmod +x
[...]

PR #528, opened 2024-02-23
fix: makeBinaryExecutable explicit filenames

edkerk:
- Fix:
  - `checkInstallation` correctly makes binaries executable on Ubuntu (solves #527)

- [x] Selected `develop` as a target branch

### DIFF
--- a/installation/checkInstallation.m
+++ b/installation/checkInstallation.m
@@ -29,6 +29,7 @@ end
 %Get the RAVEN path
 [ST, I]=dbstack('-completenames');
 [ravenDir,~,~]=fileparts(fileparts(ST(I).file));
+    status = makeBinaryExecutable();
 
 % Do not print first few lines if only version should be reported
 if ~versionOnly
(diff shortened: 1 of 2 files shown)


installation/checkInstallation.m
```diff
@@ -317,23 +318,22 @@
...
+           fullfile(binDir,'hmmer','hmmsearch');                fullfile(binDir,'hmmer','hmmsearch.mac');
+           fullfile(binDir,'GLPKmex','glpkcc.mexa64');          fullfile(binDir,'GLPKmex','glpkcc.mexglx');         fullfile(binDir,'GLPKmex','glpkcc.mexmaci64');
+           fullfile(binDir,'libSBML','TranslateSBML.mexa64');   fullfile(binDir,'libSBML','TranslateSBML.mexglx');  fullfile(binDir,'libSBML','TranslateSBML.mexmaci64');
+           fullfile(binDir,'libSBML','OutputSBML.mexa64');      fullfile(binDir,'libSBML','OutputSBML.mexglx');     fullfile(binDir,'libSBML','OutputSBML.mexmaci64');
+           fullfile(binDir,'mafft','mafft-linux64','mafft.bat');
+           fullfile(binDir,'mafft','mafft-mac','mafft.bat');};
 
 for i=1:numel(binList)
-    [status,cmdout] = system(['chmod +x "' binList{i} '".(mexa64|mexglx|mexmaci64|mac|bat)']);
+    [status,cmdout] = system(['chmod +x "' binList{i} '"']);
```
mihai-sysbio, line 336: To me it looks like the meaningful change here is the position of the second `"`, which before was before the `.` and now is at the end. I'm also a bit unsure if quotation marks were needed at all - maybe it was a mechanism to deal with any spaces that would be found in the full path of `binDir`.

Issue #529, opened 2024-02-26
bug: simplifyModel with irreversible backwards-only reactions


edkerk:
### Description of the issue:
Reactions that are defined backwards-only (so lb<0 and ub=0) could arguably have their `rev` field as `0`. But this raises a problem when `simplifyModel` is run with `deleteInaccessible` is set as `true`. In this line:

https://github.com/SysBioChalmers/RAVEN/blob/e296b5dc7b66f11d8a494303629256ec69679868/core/simplifyModel.m#L129

The S-matrix is expanded with inverted versions of reversible reactions (the `reducedModel.rev~=0` part), to consider that reactions can have a negative flux. But for the above case (backwards-only & irreversible), these reactions are not included. Arguably, it is not interesting anyway what the value of the `rev` field is, what is important is whether the reaction can carry a negative flux.

This should therefore be changed to:

`revS=[reducedModel.S,reducedModel.S(:,reducedModel.lb<0)*-1 I(:,in) I(:,out)*-1];`

**I hereby confirm that I have:**

- [x] Checked that a similar issue does not [already exist](https://github.com/SysBioChalmers/RAVEN/issues?q=is%3Aissue) or is already [discussed](https://github.com/SysBioChalmers/RAVEN/discussions).

PR #530, opened 2024-02-26
fix: various minor fixes

edkerk:
- fix:
  - `simplifyModel` with irreversible backwards-only reactions (solves #529)
  - `writeYAMLmodel` do not write lines with empty entries (e.g. reactions without subsystems)
  - `getModelFromKEGG` includes `model.annotation.defaultLB` and `model.annotation.defaultUB` fields
  - `getGenesFromGrRules` can handle genes with '|'
  - `getModelFromHomology` remove geneFrom field (solves #533)
  - `getMinNrFluxes` reduce default verbosity
  - `writeYAMLmodel` allow empty id and name fields, in line with https://github.com/SysBioChalmers/RAVEN/wiki/RAVEN-Model-Structure
  - `ravenCobraWrapper` prefers to use grRules in COBRA models if present (solves GECKO issue [#367](https://github.com/SysBioChalmers/GECKO/discussions/367))
  - `optimizeProb` will throw error when trying to solve MILP with glpk (also if glpk is set via cobra)
  - `writeYAMLmodel` will throw an informative error if it cannot write the file to the intended directory
  - `mapCompartments` correct horizontal concatenation of cell array
- feat:
  - `randomSampling` can run in parallel with MATLAB Parallel Computing Toolbox installed
  - `setParam` has an additional option, 'unc' if a reaction's lower and upper bound should be set as unconstrained. If available, this will use the default bounds in `model.annotation` (otherwise [-1000, 1000]), and considers reversibility (in which case lower bound = 0)
  - auxiliary `parallelPoolRAVEN` function to check if function should be running in parallel
  - use alternative [ProgressBar](https://github.com/elgar328/matlab-code-examples/tree/main/tools/ProgressBar), particularly in functions containing `parfor` calls
  - give execution rights in Terminal for new RAVEN functions

- [ ] This PR has `main` as target branch, and will be resolved with a *merge commit*.


edkerk:
Failing tests is likely due to invalid gurobi license on the CI server.

PR #539, opened 2024-05-02
fix: skip tests if solver is absent & glpk timeout

edkerk:
- Refactor:
  - when running a unit test and a solver is not installed, report the absence of the solver, not the general failing of the test.
- Fix:
  - prevent `glpk` timeout when running selected large FBA calculations.

- [ ] This PR has `main` as target branch, and will be resolved with a *merge commit*.


mihai-sysbio:
I'm surprised, the SCIP test should have been skipped but the test still shows it as an error.

edkerk:
No, that test explicitly is not skipped, so that we can see that SCIP is really not working. But as SCIP is already tested in that test, there is no reason to run other SCIP-based tests.

testing/unit_tests/fillGapsLargeTests.m
```diff
@@ -2,12 +2,27 @@
...
+if ~testGurobi
+    disp('Gurobi not installed or cannot be found in MATLAB path, some fillGapsLargeTests skipped.')
+    skipTests = contains({tests.Name},'gurobi','IgnoreCase',true);
+    tests(skipTests) = [];
+end
+if exist('scip','file')~=3
+    disp('SCIP MEX binary not installed or not functional, some fillGapsLargeTests skipped.')
+    skipTests = contains({tests.Name},'scip','IgnoreCase',true);
+    tests(skipTests) = [];
+end
```
mihai-sysbio, line 22: I thought the code above was meant to avoid running the SCIP test, which is then to my understanding conflicting with

> Error occurred in solverTests/testSCIP and it did not run to completion.
    ---------
    Error ID:
    ---------
    ''
    --------------
    Error Details:
    --------------
    Error using solverTests>testSCIP (line 77)
    SCIP MEX binary not installed or not functional.
..

If `SCIP MEX binary not installed` all SCIP testing should be skipped, since , the LP solver check already reports `> scip Fail`

Issue #556, opened 2024-09-13
checkModelStruct should compare .genes and .grRules


edkerk:
### Description of the issue:
If .grRules contains genes that are not in the .genes field, `exportModel` has no problem exporting the model, but the generated SBML is invalid and will cause errors when loading again. `checkModelStruct` should be the function to perform this check, as it runs at the beginning of `exportModel`.

#### Reproducing this issue:

The following code should throw an error but it doesn't.
```matlab
model2 = removeGenes(model,1);
model2 = removeUnusedGenes(model2);
model2.grRules = model.grRules;
checkModelStruct(model2);
```

#### System information
```matlab
 > Installation type                    Advanced (via git)
 > Installing from location             C:\Work\GitHub\RAVEN
 > Checking RAVEN release               2.9.3
   > You are running the latest RAVEN release
 > Checking MATLAB release              2023b
 > Checking system architecture         win64
 > Set RAVEN in MATLAB path             Pass
 > Save MATLAB path                     Pass
```

**I hereby confirm that I have:**

- [x] Checked that a similar issue does not [already exist](https://github.com/SysBioChalmers/RAVEN/issues?q=is%3Aissue) or is already [discussed](https://github.com/SysBioChalmers/RAVEN/discussions).

edkerk:
Resolved in 15b46e9ca40450af685db11235616d0a5c2db900

Issue #555, opened 2024-08-30
getExchangeRxns.m works inconsistently


dimwas:
### Description of the issue:
When searching for inflow reactions, only irreversible inflow reactions are included in the output. When looking for outflow reaction, both irreversible outflow and reversible reactions are shown.

#### Reproducing this issue:
```
load('iMK1208+suppInfo.mat') % the model provided as a part of the tutorial;
[exchangeRxns, exchangeRxnsIndexes]=getExchangeRxns(iMK1208,'both'); % 215 exchange reactions are found
[exchangeRxns, exchangeRxnsIndexes]=getExchangeRxns(iMK1208,'in'); % 0 irreversible inflow reactions are found
[exchangeRxns, exchangeRxnsIndexes]=getExchangeRxns(iMK1208,'out'); % 215 outflow reactions are found, but they are all reversible
```

#### System information
`*** THE RAVEN TOOLBOX ***

 > Installing from location             F:\RAVEN
 > Checking RAVEN release               2.8.7
   > Latest RAVEN release available     2.9.3
     Run git pull in your favourite git client
     to get the latest RAVEN release
 > Checking MATLAB release              2023a
 > Checking system architecture         win64
 > Set RAVEN in MATLAB path             Pass
 > Save MATLAB path                     Pass

=== Model import and export ===
 > Add Java paths for Excel format      Pass
 > Checking libSBML version             5.19.0
 > Checking model import and export
   > Import Excel format                Pass
   > Export Excel format                Pass
   > Import SBML format                 Pass
   > Export SBML format                 Pass

=== Model solvers ===
 > Checking for LP solvers
   > glpk                               Pass
   > gurobi                             Pass
   > scip                               Fail
   > cobra                              Pass
 > Set RAVEN solver                     gurobi

=== Essential binary executables ===
 > Checking BLAST+                      Pass
 > Checking DIAMOND                     Pass
 > Checking HMMER                       Pass

=== Compatibility ===
 > Checking function uniqueness         Pass

*** checkInstallation complete ***

ans =

    '2.8.7'`

**I hereby confirm that I have:**
- [X] Checked that a similar issue does not [already exist](https://github.com/SysBioChalmers/RAVEN/issues?q=is%3Aissue) or is already [discussed](https://github.com/SysBioChalmers/RAVEN/discussions).

edkerk:
Thank you for reporting this and the clear instructions on how to replicate. We'll work on a fix shortly.

[...]

PR #567, opened 2024-10-13
RAVEN 2.10.0

edkerk:
- Fixes:
  - `importModel` only removes identifier prefixes (like `R_` for reactions) when present in all entries of a certain type; gives a warning that this happened; and has a `removePrefix` flag that can be set to `false` if desired
  - `exportModel` only adds identifier prefixes (like `R_` for reactions) when one or more entries of a certain type are not compatible with SBML (i.e. do not start with a letter or _); gives a warning that this happened; and has a `neverPrefix` flag that can be set to `true` if desired. Previously, prefixes were always added
  - `importModel` allows for empty parameters when not specified
  - `constructS` gives correct error message when a metabolite is missing
  - `importExcelModel` can ignore faulty MIRIAM entries (note: import of Excel models is strongly discouraged!)
  - `checkModelStruct` to determine if reaction is reversible, consider all combinations of LB and UB, not just whether LB < 0 or not
  - `checkModelStruct` checks that all genes in `grRules` field are also in `genes` field (solves #556)
- Features:
  - `checkModelStruct` queries if the model has any identifiers that are not SBML compatible (i.e. do not start with a letter or _)
  - new `removeIdentifierPrefix` function can remove identifier prefixes like `R_` for reactions if desired
  - new `addIdentifierPrefix` function can add identifier prefixes like `R_` for reactions if required
  - support for `model.proteins` as field, which matches `model.genes` and is also supported by SBML and YAML
  - `getExchangeRxns` has more options to select reactions based on their reaction bounds, whether they represent uptake, excretion etc. (solves #555)
- Documentation:
  - `getKEGGModelForOrganism` explicitly refers to the correct `dataDir` options (i.e. `prok90_kegg105` or `euk90_kegg105` at this time)

- [ ] This PR has `develop` as target branch, and will be resolved with a *squash-merge*.
- [x] This PR has `main` as target branch, and will be resolved with a *merge commit*.


edkerk:
Minor release 2.9.3 -> 2.10.0, as changes in handling of identifier prefixes by `importModel` and `exportModel` can cause some changes in SBML files in comparison to earlier versions.

Issue #669, opened 2026-07-05
Possible gene ID–gene symbol mismatch in ftINIT-reconstructed models


mfcesur:
Hi,

I generated cell-specific metabolic models with ftINIT by setting "removeGenes = true".  The reconstructed models appear to contain incorrect mappings between Ensembl gene IDs (model.genes) and gene symbols (model.geneShortNames). 

To investigate this, I compared the reconstructed adipocyte model against both the original Human2 model and the Ensembl (release 109) annotations. Human2 includes the correct gene symbols, whereas the adipocyte model shows what looks like a systematic shift of gene symbols after a certain point in the gene list. Please find the attached table. 

I did not have the opportunity to investigate this in detail, but I suspect it may be related to the way gene-associated fields are updated after gene removal. This is because, _removeLowScoreGenes_ rebuilds the gene list using _getGenesFromGrRules_. Since _getGenesFromGrRules_ reconstructs the gene list using the unique function (without the 'stable' option), I was wondering whether this could change the original order of "model.genes":

`genes = unique(transpose([rxnGenes{nonEmpty}]));`

If so, geneShortNames may potentially become misaligned. I would appreciate any thoughts on whether this could explain the behavior I observed, or whether there might be another step in the ftINIT pipeline where this mismatch could be introduced.

Thank you very much for your time.

Kind regards,
Müberra

[geneSymbol_comparisons.xlsx](https://github.com/user-attachments/files/29681518/geneSymbol_comparisons.xlsx)

edkerk:
Hi Müberra,

Thank you for the detailed report — you correctly identified the bug. In `removeLowScoreGenes`, `model.genes` was regenerated in sorted order via `getGenesFromGrRules`, while `geneShortNames` (and the other gene-associated fields) were trimmed in the original gene order, causing the misalignment you observed.

This is now fixed in #670, which retains the genes in their original order. It will be included in the next RAVEN release soon. In the meantime, it would be great if you could try the updated code to confirm it resolves the mismatch on your side.

Kind regards

PR #670, opened 2026-07-07
fix: gene field alignment in removeLowScoreGenes

edkerk:
- fix:
  - `removeLowScoreGenes` misaligned the gene-associated fields (`geneShortNames`, `proteins`, `geneMiriams`, `geneFrom`, `geneComps`) relative to `model.genes`, corrupting the gene ID → gene symbol mapping in ftINIT-reconstructed models when `removeGenes = true` (solves #669).

#### Background
`removeLowScoreGenes` regenerated `model.genes` from `getGenesFromGrRules`, which returns a **sorted** list (`unique` without `'stable'`), but trimmed the gene-associated annotation fields using `remInd`, a logical mask in the **original** `model.genes` order. When `model.genes` was not already sorted (as for Human-GEM–derived models), the annotation fields ended up shifted relative to `model.genes` — exactly the "systematic shift of gene symbols after a certain point" reported in #669. GPRs and `rxnGeneMat` stayed internally consistent, so only the ID↔symbol mapping was affected.

The defect has been present since `removeLowScoreGenes.m` was first moved into RAVEN (May 2022, commit d7b8ba2b).

#### Fix
Gene removal never introduces new genes, so the retained set is `model.genes(~remInd)`. Keep that original order for `newModel.genes` and reorder the `rxnGeneMat` columns to match, so the `remInd`-trimmed annotation fields stay aligned.

- [X] This PR has `develop` as target branch, and will be resolved with a *squash-merge*.
- [ ] This PR has `main` as target branch, and will be resolved as descriped [here](https://github.com/SysBioChalmers/RAVEN/wiki/Development-policy#make-a-new-release).


mfcesur:
Hi, 

Thank you very much for the quick fix. 

I tested the latest version from the develop branch, and the gene ID-gene symbol mapping are now correct. The issue seems to be resolved on my end. 

Kind regards, 
Müberra

